PR #2, opened 2022-04-19
Datacite & JPAL transform

ehanson8:
#### Helpful background context
- In general, I’m using `hoard` as a model and copied some validator functions over. 
- For some functions, I thought the naming and description of the args was explicit enough that a descriptive docstring would be redundant but I’m happy to add them for any that need to be clearer.
- I've included the JPAL records Helen harvested as a fixture but I want to clarify in the PR discussion how we should utilize those

#### How can a reviewer manually see the effects of these changes?
I'm not sure how we want to do manual testing for this repo going forward so I hope we can hash that out through the PR discussion. I'm assuming Helen's work on the OAI Harvester will be involved. 

#### What are the relevant tickets?
https://mitlibraries.atlassian.net/browse/RDI-74

#### Developer
- [x] All new ENV is documented in README
- [x] Stakeholder approval has been confirmed (or is not needed)

#### Code Reviewer
- [ ] The commit message is clear and follows our guidelines
      (not just this pull request message)
- [ ] There are appropriate tests covering any new functionality
- [ ] The documentation has been updated or is unnecessary
- [ ] The changes have been verified
- [ ] New dependencies are appropriate or there were no changes

#### Includes new or updated dependencies?
YES

PR #3, opened 2022-04-27
Updated transform

ehanson8:
Building off on @hakbailey's framework, this PR replaces #2

Let's discuss how we want to handle this going forward

https://mitlibraries.atlassian.net/browse/RDI-74

#### Developer

#### Code Reviewer

YES


tests/test_datacite.py
```diff
@@ -1,22 +1,119 @@
...
-# missing required Datacite fields raises error
-# all fields transformed successfully
-# repeatable fields transformed successfully
+from transmogrifier.models import Contributor, Date, Identifier, Note, TimdexRecord
+from transmogrifier.sources.datacite import Datacite
 
 
-def test_jpal_fixtures():
-    input_records = parse_xml_records("tests/fixtures/datacite_jpal_records.xml")
+def test_jpal_fixtures(datacite_jpal_records):
```
hakbailey, line 7: This was my bad but we should name this better, something along the lines of `test_datacite_iterates_through_all_records`.

tests/test_datacite.py
```diff
@@ -1,22 +1,119 @@
...
         "https://dataverse.harvard.edu/dataset.xhtml?persistentId=",
         "Abdul Latif Jameel Poverty Action Lab Dataverse",
-        input_records,
+        datacite_jpal_records,
     )
     assert len(list(output_records)) == 38
 
 
-# TODO: add tests for generate_name_identifier_url expected cases
+def test_datacite_full_record(datacite_record, datacite_jpal_record_full):
```
hakbailey, line 17: I would use the same language here (and in conftest.py) that we use in the models tests, e.g. `datacite_record_all_fields`, `datacite_record_timdex_required_fields`, etc. as I think it's more clear than full and minimal.

tests/test_datacite.py
```diff
@@ -1,22 +1,119 @@
...
+    datacite_record.input_records = datacite_jpal_record_full
+    output_records = datacite_record
+    for output_record in output_records:
```
hakbailey, line 20: Since we only expect a single output record, this will be more clear if we just use `assert next(output_records) == ...` (same for all tests below).

tests/fixtures/datacite_jpal_record_full.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hakbailey, line 1: I would consider putting these fixtures in a "datacite" subfolder and renaming them to just "jpal_record...". I think it will be more clear if the fixtures subfolders match our sources folder (especially as we add potentially a lot more source transforms over time).

hakbailey (changes requested):
In general this looks good, see inline comments for requested changes. I would also update the readme to reflect the new single command and env. Happy to chat about any of this! (And also about the "How can a reviewer manually see the effects of these changes?" question.)

transmogrifier/sources/datacite.py
```diff
@@ -8,12 +8,12 @@
 class Datacite:
     def __init__(
         self,
-        source: str,
+        source_symbol: str,
```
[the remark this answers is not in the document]
ehanson8: I can change this back but we have the Datacite `source_name` going to the `source` kwarg of the TIMDEX class and Datacite `source` is only used to prepend the ID in the `source_link_url`. We can see this now but I think this is easily avoidable confusion for when we or someone else returns to this code later. We could also change TIMDEX's `source` to `source_name` for consistency on that end at least

Makefile
```diff
@@ -22,4 +22,4 @@ mypy:
 	pipenv run mypy transmogrifier
 
 test:
-	pipenv run pytest --cov=transmogrifier
\ No newline at end of file
+	pipenv run pytest --cov-report term-missing --cov=transmogrifier
```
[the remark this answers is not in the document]
ehanson8: Sounds good

tests/test_datacite.py
```diff
@@ -14,106 +14,110 @@ def test_jpal_fixtures(datacite_jpal_records):
...
 ):
     with raises(ValueError):
-        datacite_record.input_records = datacite_jpal_record_multiple_titles
-        output_records = datacite_record
-        for output_record in output_records:
-            assert output_record
+        output_records = datacite_record_partial(
+            input_records=datacite_jpal_record_multiple_titles
+        )
+        assert next(output_records)
```
hakbailey, line 103: This assert and the one in the above test aren't necessary since all these are testing is that the error was raised.

tests/test_datacite.py
```diff
@@ -14,106 +14,110 @@ def test_jpal_fixtures(datacite_jpal_records):
...
-        output_records = datacite_record
-        for output_record in output_records:
-            assert output_record
+        output_records = datacite_record_partial(
+            input_records=datacite_jpal_record_missing_required_field
+        )
+        assert next(output_records)
 
 
 def test_datacite_multiple_titles(
```
hakbailey, line 96: These tests should also be named to indicate what they are testing (raises error).

hakbailey (approved):
See a couple of inline suggestions to optionally clarify test names/functionality, but this looks good.

PR #5, opened 2022-05-04
RDI-77

ehanson8:
This PR completes the initial fields that were mapped for JPAL. Certain attributes (e.g `Location.geodata`) may be added later as relevant examples are found in the JPAL records.

I'll reach out about how we want to do this when I get back but I wanted to get this PR out before my day off.

https://mitlibraries.atlassian.net/browse/RDI-77

#### Developer

#### Code Reviewer
- [x] The changes have been verified

NO


transmogrifier/sources/datacite.py
```diff
@@ -117,6 +155,43 @@ def create_from_datacite_xml(
...
+                    f.funder_identifier_type = f_id_type
+            if funding_reference.find("awardNumber"):
+                award_number = funding_reference.find("awardNumber")
+                f.award_number = award_number.string
+                if "awardURI" in award_number.attrs:
+                    f.award_uri = award_number.attrs["awardURI"]
+            kwargs.setdefault("funding_information", []).append(f)
 
         # identifiers
         identifier_xml = xml.metadata.find("identifier")
```
ehanson8, line 197: Taking a 2nd look at some of the code from the previous PR, is there potentially an issue in grabbing `xml.metadata.find("identifier")` here and grabbing `xml.header.find("identifier")` for `source_record_id`?

tests/test_models.py
```diff
@@ -1,45 +1,92 @@
...
+    assert timdex_record_required_fields.summary is None
 
 
 def test_timdex_record_required_subfields_only(timdex_record_required_fields):
     timdex_record_required_fields.contributors = [Contributor(value="Smith, Jane")]
     timdex_record_required_fields.identifiers = [Identifier(value="123")]
-    timdex_record_required_fields.notes = [Note(value=["This book is awesome"])]
+    timdex_record_required_fields.notes = [Note(value="This book is awesome")]
+    timdex_record_required_fields.alternate_titles = [AlternateTitle(value="Alt Title")]
+    timdex_record_required_fields.locations = [
```
hakbailey, line 45: Location check can be removed.

transmogrifier/sources/datacite.py
```diff
@@ -126,6 +201,27 @@ def create_from_datacite_xml(
...
+        # locations
+        locations = xml.metadata.find_all("geoLocationPlace")
+        for location in locations:
+            kwargs.setdefault("locations", []).append(Location(value=location.string))
+
+        # notes
+        descriptions = xml.metadata.find_all("description")
+        for description in [
+            d
+            for d in descriptions
```
hakbailey, line 218: `descriptionType` attribute is mandatory in Datacite if the `Description` field is used, so we should do here what we do for other mandatory fields and log a warning if it's not present.

transmogrifier/sources/datacite.py
```diff
@@ -137,6 +233,40 @@ def create_from_datacite_xml(
         else:
             kwargs["publication_information"] = [publisher.string]
 
+        # related_items
+        related_items = xml.metadata.find_all("relatedIdentifier")
+        for related_item in related_items:
```
hakbailey, line 238: It looks like we need to use the `relatedIdentifierType` to construct a URL here like we do for the name identifiers in Creator and Contributor (and probably by calling a very similar function).

tests/conftest.py
```diff
@@ -67,15 +73,15 @@ def datacite_record_partial():
         Datacite,
         source="cool-repo",
         source_name="A Cool Repository",
-        source_base_url="https://example.com/item123",
+        source_base_url="https://example.com/",
```
hakbailey, line 76: This change to the Datacite fixture makes sense 👍

tests/test_datacite.py
```diff
@@ -56,10 +71,76 @@ def test_datacite_record_all_fields(
...
+            "application/pdf",
+        ],
+        format="electronic resource",
+        funding_information=[
+            Funder(
+                funder_name="3ie, Nike Foundation",
+                funder_identifier="0987",
+                funder_identifier_type="Crossref FunderID",
+                award_number="OW1/1012 (3ie)",
+                award_uri="7689",
```
hakbailey, line 105: Since we're making this up can we make it look like a real URI for clarity?

hakbailey (changes requested):
Most of this looks great, see inline comments for specific change requests. There are a couple of changes that should probably be new tickets instead of doing them as part of this one, would you mind adding those to the project backlog? Thanks!

transmogrifier/models.py
```diff
@@ -69,15 +75,60 @@ class Date:
...
+class Rights:
+    description: Optional[str] = field(
+        default=None, validator=optional(instance_of(str))
+    )
+    uri: Optional[str] = field(default=None, validator=optional(instance_of(str)))
+
+
+@define
+class Subject:
+    value: str = field(validator=instance_of(str))  # Required subfield
```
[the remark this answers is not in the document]
ehanson8: Same concern here as the `Note`'s value field, also, doesn't collapsing all the `Subject `values together compromise our ability to facet by a particular subject headings?

transmogrifier/models.py
```diff
@@ -69,15 +75,60 @@ class Date:
...
+class Location:
+    value: str = field(validator=instance_of(str))  # Required subfield
+    kind: Optional[str] = field(default=None, validator=optional(instance_of(str)))
+    geodata: Optional[str] = field(default=None, validator=optional(instance_of(str)))
+
+
 @define
 class Note:
-    value: list[str] = field(validator=list_of(str))  # Required subfield
+    value: str = field(validator=instance_of(str))  # Required subfield
```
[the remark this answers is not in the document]
hakbailey: There can be multiple note objects of the same kind, that's totally fine. This isn't saying we _have_ to combine notes of the same type into one array, it's giving us the option for sources where that makes sense.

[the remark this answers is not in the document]
ehanson8: OK, concerns withdrawn!

transmogrifier/models.py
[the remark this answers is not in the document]
hakbailey: I think we can do that faceting either way, but which way is easiest is a conversation to have with the bigger team. For now, we need this PR to work with the data model as it is. Definitely add tickets to discuss data model changes where we need to do that!

transmogrifier/sources/datacite.py
```diff
@@ -47,21 +59,47 @@ def create_from_datacite_xml(
         """
         # Required fields in TIMDEX
         source_record_id = xml.header.find("identifier").string
+        kwargs = {
+            "source": source_name,
+            "source_link": source_link_url + source_record_id,
+            "timdex_record_id": f"{source}:{source_record_id.replace('/', '-')}",
+            "format": "electronic resource",
```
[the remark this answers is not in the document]
ehanson8: My intent was to insert a constant at the creation of the `kwargs` but I see your point on the logical flow of the code

transmogrifier/sources/datacite.py
```diff
@@ -47,21 +59,47 @@ def create_from_datacite_xml(
...
                 "A record must have exactly one title. Titles found for record "
                 f"{source_record_id}: {main_title}"
             )
-        kwargs = {
-            "source": source_name,
-            "source_link": source_link_url + source_record_id,
-            "timdex_record_id": f"{source}:{source_record_id.replace('/', '-')}",
-            "title": main_title[0].string,
-        }
+        subtitles = [
```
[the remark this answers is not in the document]
ehanson8: That cleans this up considerably, good call

tests/conftest.py
```diff
@@ -80,8 +101,9 @@ def datacite_record_partial():
 @pytest.fixture()
 def timdex_record_required_fields():
     return TimdexRecord(
+        citation="Creator (PubYear): Title. Publisher. (resourceTypeGeneral). ID",
```
hakbailey, line 104: See note about citation, we shouldn't include it in the required TIMDEX fields record.

tests/test_datacite.py
```diff
@@ -80,6 +87,12 @@ def test_datacite_record_all_fields(
                 range=None,
                 value="2019-06-24",
             ),
+            Date(
+                kind="Collected",
+                note=None,
+                range=Date_Range(gt=None, gte="2007-01-0", lt=None, lte="2007-02-28"),
```
hakbailey, line 93: This test shows where we're losing the last digit of the `gte` date.

hakbailey (changes requested):
Very close, just a couple more changes and this will be good to go. Thanks!

transmogrifier/sources/datacite.py
```diff
@@ -152,3 +299,17 @@ def generate_name_identifier_url(cls, name_identifier):
...
+
+    @classmethod
+    def generate_related_item_identifier_url(cls, related_item_identifier):
+        """
+        Generate a full related item identifier URL with the specified scheme.
+        Args:
+            related_item_identifier: An BeautifulSoup Tag Tag representing a Datacite
+            relatedIdentifier XML field.
+        """
+        if related_item_identifier["relatedIdentifierType"] == "DOI":
```
hakbailey, line 311: Caught a bug during verification. Both of these generate identifier url functions need to use `dict.get(key)` to account for cases where there is no `IdentifierType` attribute. We should add tests for that too.

ehanson8: Good catch, fixed and added tests!

hakbailey (approved):
Sweet!

PR #7, opened 2022-06-02
makefile and workflow updates

zotoMIT:
# Subject
Applies a universal github workflow and standard makefile commands to the app.

# Why these changes are being introduced:
This commit is part of a larger refactor of how we handle ECR repositories and generating makefile and github workflows.
The work here will likely be removed when stage and prod are deployed, so this is a temporary change, but serves as a proof of concept/stopgap until stage is ready.
The changes here represent a direct copy/paste from the infrastructure repository.

# How this addresses that need:
Add the dev-build workflow
Update the makefile with our new standard format

# Side effects of this change:
A test of this workflow will happen when I submit the PR for this.

# Relevant ticket(s):
NA


ehanson8 (approved):
This looks like a great approach

PR #10, opened 2022-06-09
RDI-55-whoas-transform

ehanson8:
As I noted in the mapping spreadsheet, the `dc.description.sponsorship` value appears to always be a descriptive statement and the `Funder.funder_name` attribute is the best available option but not a great fit. Not sure if this justifies creating a `description` attribute for the `Funder` class but just noting it here as well.

Pending discussion of how we want to do this

* https://mitlibraries.atlassian.net/browse/RDI-55

#### Developer

#### Code Reviewer

NO


transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,276 @@
...
+            source_base_url: The base URL for the source system from which direct links
+            to source metadata records can be constructed.
+            source_name: The full human-readable name of the source repository to be
+            used in the TIMDEX record.
+            xml: A BeautifulSoup Tag representing a single DSpace record in
+            dim XML.
+        """
+        # Required fields in TIMDEX
+        source_record_id = cls.create_source_record_id(xml)
+        all_fields = xml.metadata.find_all("field")
```
hakbailey, line 62: Rather than finding all fields and then searching within those, we can just search directly for the fields we need like we do in other classes, using bs4's option to pass an attribute into the `find_all` method. So titles, for example, would look something like:
```
all_titles= xml.find_all("dim:field", element="title")
```

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,276 @@
...
+        ]:
+            a = AlternateTitle(
+                value=alternate_title.string,
+                kind="AlternativeTitle",
+            )
+            kwargs.setdefault("alternate_titles", []).append(a)
+
+        # contributors
+        citation_creators = []
+        for creator in [f for f in all_fields if f.attrs["element"] == "creator"]:
```
hakbailey, line 92: We can just pull all contributors here and use the qualifier value as the kind value. Also, all of the records in the whoas records fixture seem to use "author" as the qualifier for authors so we probably want to use that to identify citation contributors.

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,276 @@
...
+        ]:
+            fr = Funder(
+                funder_name=funding_reference.string,
+            )
+            kwargs.setdefault("funding_information", []).append(fr)
+
+        # identifiers
+        identifiers = [f for f in all_fields if f.attrs["element"] == "identifier"]
+        for identifier in identifiers:
+            if "qualifier" not in identifier.attrs:
```
hakbailey, line 150: This identifiers section can be simplified by just excluding citation identifiers and then using the dict.get() method to set the kind, e.g.:
```
if identifier.get("qualifier") != "citation":
    i = Identifier(
        value=identifier.string,
        kind=identifier.get("qualifier"),
    )
```
There are several other fields below where a similar simplification can be made.

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,276 @@
...
+                )
+                kwargs.setdefault("identifiers", []).append(i)
+
+        # language
+        for language in [
+            f
+            for f in all_fields
+            if f.attrs["element"] == "language"
+            and "qualifier" in f.attrs
+            and f.attrs["qualifier"] == "iso"
```
hakbailey, line 171: Is there a reason we're restricting to languages with the "iso" qualifier?

hakbailey, line 171: Is there a reason we're explicitly restricting to languages with the "iso" qualifier?

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,276 @@
...
+                    kind=description.attrs["qualifier"],
+                )
+                kwargs.setdefault("notes", []).append(n)
+
+        # publication_information
+        for publisher in [f for f in all_fields if f.attrs["element"] == "publisher"]:
+            kwargs.setdefault("publication_information", []).append(publisher.string)
+
+        # related_items, uses related_identifiers retrieved for identifiers
+        for related_item in [f for f in all_fields if f.attrs["element"] == "relation"]:
```
hakbailey, line 201: Same simplification here as for identifiers.

hakbailey (changes requested):
See initial comments on the dspace_dim module. I would also encourage you to think about what happens if any of the string values of these fields are empty (which I discovered in the records I harvested is definitely a possibility in DSpace). I would add tests for that and the code will likely need to be updated to handle it.

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,276 @@
...
+            "timdex_record_id": f"{source}:{source_record_id.replace('/', '-')}",
+            "title": main_title[0].string,
+        }
+
+        # Optional fields in TIMDEX
+        # alternate_titles, uses full title list retrieved for main title field
+        for alternate_title in [
+            t
+            for t in all_titles
+            if "qualifier" in t.attrs and t.attrs["qualifier"] == "alternative"
```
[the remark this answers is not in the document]
ehanson8: This was an attempt to normalize the qualifier with the value from Datacite but rethinking it, that should be handled more globally and consistently across other fields as well and perhaps not even in this app. I'll change it

ehanson8:
I added code and tests to account for blanks fields, it's not anything I've encountered with this collection so far but you're right that it's good to be prepared for that

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,243 @@
...
+        for contributor in [
+            c for c in xml.find_all("dim:field", element="contributor") if c.string
+        ]:
+            if contributor.get("qualifier") == "author":
+                citation_creators.append(contributor.string)
+            c = Contributor(value=contributor.string, kind=contributor.get("qualifier"))
+            kwargs.setdefault("contributors", []).append(c)
+
+        # dates
+        date_issued = ""
```
hakbailey, line 108: We should also give `date_issued` a kind of `Publication date` to match the other sources.

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,243 @@
...
+            d for d in descriptions if d.get("qualifier") == "abstract" and d.string
+        ]:
+            kwargs.setdefault("summary", []).append(description.string)
+
+        # citation, uses identifiers list retrieved for identifiers field
+        citation = ""
+        for identifier in [
+            i for i in identifiers if i.get("qualifier") == "citation" and i.string
+        ]:
+            citation = identifier.string
```
hakbailey, line 231: This is going to result in `citation` being set to the last citation value if there are multiple (which there shouldn't be, but...). Since TIMDEX only allows one citation string, I would probably go with the first one? e.g. `xml.find("dim:identifier", qualifier="citation")`.

tests/test_dspace_dim.py
```diff
@@ -0,0 +1,294 @@
...
+    )
+    assert next(output_records) == TimdexRecord(
+        source="A Cool Repository",
+        source_link="https://example.com/oai:darchive.mblwhoilibrary.org:1912/2641",
+        timdex_record_id="cool-repo:oai:darchive.mblwhoilibrary.org:1912-2641",
+        title=(
+            "Time lapse movie of meiosis I in a living spermatocyte from the crane "
+            "fly, Nephrotoma suturalis, viewed with polarized light microscopy"
+        ),
+        alternate_titles=None,
```
hakbailey, line 183: Not a big deal, but we don't really need to put all these `None` arguments in since that's the default for all non-required fields.

tests/test_dspace_dim.py
```diff
@@ -0,0 +1,294 @@
...
+            " https://example.com/oai:darchive.mblwhoilibrary.org:1912/2641"
+        ),
+        content_type=None,
+        contributors=None,
+        dates=[Date(kind="issued", note=None, range=None, value="2002-11")],
+        edition=None,
+        file_formats=None,
+        format="electronic resource",
+        funding_information=None,
+        identifiers=[Identifier(value="https://hdl.handle.net/1912/2641", kind="uri")],
```
hakbailey, line 196: This test should just include the required fields plus citation since that always gets generated.

tests/test_dspace_dim.py
```diff
@@ -0,0 +1,294 @@
...
+    dspace_dim_record_partial, dspace_dim_record_no_title
+):
+    with raises(ValueError):
+        output_records = dspace_dim_record_partial(
+            input_records=dspace_dim_record_no_title
+        )
+        next(output_records)
+
+
+def test_dspace_dim_record_unqualified_identifier(
```
hakbailey, line 270: I'm not sure I understand the purpose of this test?

hakbailey (changes requested):
See inline comments. Also, all of the new fixtures should not be in the datacite folder. I'd suggest moving them to a new dspace folder in fixtures since the dspace mets fixtures will also go there.

tests/test_dspace_dim.py
ehanson8: It is unnecessary after the last round of changes

tests/test_dspace_dim.py
```diff
@@ -0,0 +1,294 @@
...
+def test_dspace_dim_record_all_fields(
+    dspace_dim_record_partial, dspace_dim_record_all_fields
+):
+    output_records = dspace_dim_record_partial(
+        input_records=dspace_dim_record_all_fields
+    )
+    assert next(output_records) == TimdexRecord(
+        citation="Journal of Geophysical Research: Solid Earth 121 (2016): 5859–5879",
+        source="A Cool Repository",
+        source_link="https://example.com/oai:darchive.mblwhoilibrary.org:1912/2641",
```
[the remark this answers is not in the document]
ehanson8: I had the incorrect source link format but just confirming, does `whoas:oai:darchive.mblwhoilibrary.org:1912-2641` what you're expecting for the `timdex_record_id`?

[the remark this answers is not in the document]
hakbailey: No, I would expect something like what we did for Zenodo to remove all that extra stuff before the actual ID.

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,243 @@
...
+            kwargs.setdefault("contributors", []).append(c)
+
+        for contributor in [
+            c for c in xml.find_all("dim:field", element="contributor") if c.string
+        ]:
+            if contributor.get("qualifier") == "author":
+                citation_creators.append(contributor.string)
+            c = Contributor(value=contributor.string, kind=contributor.get("qualifier"))
+            kwargs.setdefault("contributors", []).append(c)
+
```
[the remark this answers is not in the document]
ehanson8: I've added this but I haven't found an usage of it in WHOAS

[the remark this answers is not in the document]
hakbailey: Sure, but this isn't a WHOAS transform it's a dspace dim transform that we may use on other sources. For additional context: I'm thinking we might want to run the DSpace@MIT records through this transform too and see if we get better data than the mets. It turns out a LOT of dc fields aren't mapped to the OAI-PMH mets even though they are mapped in the general mets crosswalk (go figure). But inherently all dc metadata fields are mapped to dim because it's the internal format used for dspace metadata. The catch is that some data (like file formats) will only show up in dim if it's explicitly added to the dc fields, whereas it will always show up in the mets filesec. So which metadata type we want to use may ultimately depend on which fields we think are most important for a given source. Anyway that was a lot of info and I'm adding a spike ticket to investigate this question :)

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,298 @@
...
+        kwargs["contents"] = [
+            t.string
+            for t in xml.find_all(
+                "dim:field", element="description", qualifier="tableofcontents"
+            )
+            if t.string
+        ] or None
+
+        # contributors
+        citation_creators = []
```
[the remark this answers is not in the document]
ehanson8: Cleaned up!

hakbailey (changes requested):
Some questions and suggestions, plus a few things that need to be addressed.

transmogrifier/sources/dspace_dim.py
```diff
@@ -0,0 +1,298 @@
...
+            kwargs.setdefault("contributors", []).append(
+                Contributor(value=contributor.string, kind=contributor.get("qualifier"))
+            )
+
+        # dates
+        for date in [d for d in xml.find_all("dim:field", element="date") if d.string]:
+            if date.get("qualifier") == "issued":
+                d = Date(value=date.string, kind="Publication")
+            else:
+                d = Date(value=date.string, kind=date.get("qualifier"))
```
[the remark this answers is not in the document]
ehanson8: Even though they're not formatted according to DCMI, `dc.coverage.temporal` values tend to be like `20160701 - 20160731 (UTC)` and `Tag on time: 11:41:37 Local`, so `value` seemed a bit more appropriate initially but considering that they aren't parsable, I'll go back to the original plan of putting them in `note`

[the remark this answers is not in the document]
hakbailey: Yeah, looking at the data I agree `note` makes more sense here. For reference, the `date.value` field in Opensearch gets automatically converted to a date internally when indexed _if_ it uses one of several standard date string formats (yyyy, yyyy-mm, yyyymm, yy-mm-dd, yyyymmdd, etc.). I expect that eventually we'll use those date values to limit searches, e.g. "limit results to items published after 2020" which means we should try to restrict the `value` subfield to strings that generally look like single dates...which obviously won't be perfect given source data issues, but avoiding it when we know the data doesn't match that format is probably a good idea.

tests/test_cli.py
```diff
@@ -64,6 +64,25 @@ def test_cli_dspace_mets(caplog, runner, tmp_path):
     assert "Completed transform, total record count: 12" in caplog.text
 
 
+def test_cli_whoas(caplog, monkeypatch, runner, tmp_path):
```
hakbailey, line 67: Can remove unused `monkeypatch` arg. Honestly I'm surprised (and mildly annoyed) that our linters don't catch that!

tests/fixtures/dspace/dspace_dim_record_no_citation_field.xml
```diff
@@ -0,0 +1,30 @@
+<records>
```
hakbailey, line 1: I think we can remove this fixture as it's no longer used in the tests.

hakbailey (approved):
See inline comments for two small cleanup changes, but once those are done this looks good to go!

tests/test_cli.py
ehanson8: It does show up as faded in VS Code so I should have caught it and I'll keep on eye out going forward

PR #12, opened 2022-06-16
workflow and makefile updates

zotoMIT:
Now that we have landed on a final version of workflows and makefiles for our projects, I need to do 1 more revision of the Workflows and makefile.

This sets up/refines all workflows and the makefile

It adds the prod promote workflow, which will not work until we deploy prod

This is a copy/paste from the ECR infrastructure output, this output is available to the devs in terraform cloud.


cabutlermit (approved):
Looks good.

PR #14, opened 2022-06-17
Update sentry and logging configuration

hakbailey:
#### What does this PR do?
Refactors configuration into its own module, adds some useful logging, plus a bunch of standardization/cleanup changes.

Our terraform template requires that each env variable be present in all envs, even if we don't need it in a particular env (e.g. SENTRY_DSN is not needed in dev). Env variables are now given a string value of "None" in env(s) where they are not needed.

Run a transform locally with `SENTRY_DSN=None` in .env to confirm that it configures correctly and that the logging output is as expected and includes new total transformation time information. Sample command:
```
pipenv run transform -i tests/fixtures/datacite/datacite_records.xml -o output/datacite-timdex-records.json -s jpal
```

Also verify successful GH actions run using shared workflow.

- https://mitlibraries.atlassian.net/browse/RDI-160
- https://mitlibraries.atlassian.net/browse/RDI-148

YES

#### Developer

#### Code Reviewer


ehanson8 (approved):
Looks good and works as expected, I did notice that we don't have a unit test for `write_timdex_records_to_json`, I assume `coverage` is counting that as covered by the CLI tests.  I don't think that needs to block this but we might circle back to creating that

hakbailey:
Yeah I had commented in the code initially to add that test but removed it here because tickets are better. On my to-do list to add a ticket for it :)

> Looks good and works as expected, I did notice that we don't have a unit test for `write_timdex_records_to_json`, I assume `coverage` is counting that as covered by the CLI tests. I don't think that needs to block this but we might circle back to creating that

PR #17, opened 2022-06-24
RDI-151-datacite-refactor

ehanson8:
I believe `datacite_record_required_fields` is basically the same type of fixture as the `_record_optional_fields_missing` fixtures for DSpace so I renamed it to align with those. Also, the `rights` parsing is a bit ugly so I would welcome a cleaner approach but the logic is complicated since it can be a URI or a descriptive string.

Run transform locally on the datacite_records.xml fixture

```
pipenv run transform -i tests/fixtures/datacite/datacite_records.xml -o dspace-mets-samples-timdex.json -s jpal
```

* https://mitlibraries.atlassian.net/browse/RDI-151

#### Developer

#### Code Reviewer

NO


transmogrifier/sources/datacite.py
```diff
@@ -188,59 +201,62 @@ def create_from_datacite_xml(
...
             )
             kwargs["identifiers"].append(i)
 
         # language
         language = xml.metadata.find("language")
-        if language:
+        if language and language.string:
             kwargs["languages"] = [language.string]
 
         # locations
```
hakbailey, line 231: We might consider adding the Digital object URL link here since we have it in the other two transforms. In this case I think it would be the same as the source link unless you can see a different field to pull from. (Could be a new ticket)

transmogrifier/sources/datacite.py
```diff
@@ -188,59 +201,62 @@ def create_from_datacite_xml(
...
+            gl for gl in xml.metadata.find_all("geoLocationPlace") if gl.string
+        ]:
             kwargs.setdefault("locations", []).append(
                 timdex.Location(value=location.string)
             )
 
         # notes
         descriptions = xml.metadata.find_all("description")
-        for description in descriptions:
+        for description in [d for d in descriptions if d.string]:
```
hakbailey, line 241: The `descriptions` variable is set outside the list comprehension here but inside it in most of the other cases. Was that intentional?

transmogrifier/sources/datacite.py
```diff
@@ -188,59 +201,62 @@ def create_from_datacite_xml(
...
                     "Datacite record %s missing required Datacite attribute "
                     "@descriptionType",
                     source_record_id,
                 )
             else:
-                if description.attrs["descriptionType"] != "Abstract":
-                    n = timdex.Note(
-                        value=[description.string],
-                        kind=description.attrs["descriptionType"],
+                if description.get("descriptionType") != "Abstract":
```
hakbailey, line 249: I think we should remove the `else` clause here. We want to add the note even if there's no `descriptionType`, right?

transmogrifier/sources/datacite.py
```diff
@@ -188,59 +201,62 @@ def create_from_datacite_xml(
...
+                            value=[description.string],
+                            kind=description.get("descriptionType"),
+                        )
                     )
-                    kwargs.setdefault("notes", []).append(n)
 
         # publication_information
         publisher = xml.metadata.find("publisher")
-        if publisher is None:
+        if publisher is None or publisher.string is None:
```
hakbailey, line 259: Same note as for `publication_year`.

transmogrifier/sources/datacite.py
```diff
@@ -249,23 +265,23 @@ def create_from_datacite_xml(
...
         # rights
-        rights_list = xml.metadata.find_all("rights")
-        for rights in rights_list:
-            r = timdex.Rights()
-            if rights.string:
-                r.description = rights.string
-            if "rightsURI" in rights.attrs:
-                r.uri = rights.attrs["rightsURI"]
-            kwargs.setdefault("rights", []).append(r)
+        for rights in xml.metadata.find_all("rights"):
```
hakbailey, line 277:
```
for right in [
    r for r in xml.metadata.find_all("rights") if r.string or r.get("rightsURI")
]:
    kwargs.setdefault("rights", []).append(
        timdex.Rights(
            description=right.string or None, uri=right.get("rightsURI")
        )
    )
```

transmogrifier/sources/datacite.py
```diff
@@ -249,23 +265,23 @@ def create_from_datacite_xml(
...
+            if rights.string or rights.get("rightsURI"):
+                r = timdex.Rights()
+                if rights.string:
+                    r.description = rights.string
+                if rights.get("rightsURI"):
+                    r.uri = rights.get("rightsURI")
+                kwargs.setdefault("rights", []).append(r)
 
         # subjects
         subjects_dict: Dict[str, List[str]] = {}
```
hakbailey, line 287: Could also update subjects:
```
kwargs["subjects"] = [
    timdex.Subject(value=value, kind=key)
    for key, value in subjects_dict.items()
] or None
```

transmogrifier/sources/datacite.py
```diff
@@ -287,7 +303,7 @@ def create_from_datacite_xml(
         for description in [
```
hakbailey, line 303: Why not:
```
kwargs["summary"] = [
    d.string
    for d in descriptions
    if d.get("descriptionType") == "Abstract" and d.string
] or None
```

hakbailey (changes requested):
All the comments here are super nit-picky (and thus optional) but that's kind of the point of this whole PR so... ¯\_(ツ)_/¯ 

In addition to the inline comments, I noticed as I was going through the source transforms side-by-side that there are also some formatting discrepancies between the two dspace transforms. Do we want to update those as part of this ticket?

transmogrifier/sources/datacite.py
ehanson8: The `descriptions` variable also gets used for `summary` so that's why I kept it outside of the list comprehension

transmogrifier/sources/datacite.py
ehanson8: Much appreciated, I wasn't able to see that

ehanson8:
Made the changes, great suggestions!

transmogrifier/sources/datacite.py
ehanson8: I added this but there was discussion in the retro about whether it was useful to have duplicate values here so we might discuss this with the team

hakbailey (approved):
👍

PR #20, opened 2022-07-12
RDI-166-record-filtering

ehanson8:
I was aiming for minimal changes to the type hinting but I'm of course open to a different approach for how to make `transform` return `None` based on the response from `get_content_type` method

Run the following commands to ensure that the `Zenodo` and `WHOAS` classes do not transform records with unaccepted `content_type` values and still transform records with the expected values.

```
pipenv run transform -i tests/fixtures/datacite/datacite_records.xml -o jpal.json -s zenodo
```
Output should be 38

```
pipenv run transform -i tests/fixtures/datacite/zenodo_record_unaccepted_content_type.xml -o zenodo.json -s zenodo
```
Output should be 0

```
pipenv run transform -i tests/fixtures/dspace/dspace_dim_records.xml -o dspace.json -s whoas
```
Output should be 4

```
pipenv run transform -i tests/fixtures/dspace/whoas_record_unaccepted_content_types.xml -o whoas.json -s whoas
```
Output should be 0

* https://mitlibraries.atlassian.net/browse/RDI-166

#### Developer

#### Code Reviewer

NO


hakbailey (changes requested):
This approach is a good start but it's missing a key element: what happens if a `None` value is returned while iterating through multiple input records instead of just one? Try tacking a record with valid content types onto the end of one of these new fixtures and check the output.

In addition, now that we have both transformed and skipped records, we probably want to log counts of those at the end in additional to the total number of records processed. That will give us a rough indication of whether the transform process matched our expectations.

hakbailey (changes requested):
See comments inline, this is looking good but I think some of the logic needs to be moved around for clarity and extensibility.

transmogrifier/sources/datacite.py
```diff
@@ -331,3 +337,13 @@ def generate_related_item_identifier_url(cls, related_item_identifier: Tag) -> s
         else:
             base_url = ""
         return base_url + related_item_identifier.string
+
+    @classmethod
+    def valid_content_types(cls, content_type_list: List[str]) -> bool:
+        """
```
hakbailey, line 343: I would update the docstring to clarify that this may be overridden by subclasses that require content type validation, since the method here doesn't actually do that.

transmogrifier/sources/whoas.py
```diff
@@ -0,0 +1,34 @@
+from typing import List
```
hakbailey, line 1: Totally optional, but this is unnecessary. You can just use the built-in type `list` for type hinting.

transmogrifier/sources/zenodo.py
```diff
@@ -18,3 +20,23 @@ def get_source_record_id(cls, xml: Tag) -> str:
...
+    def valid_content_types(cls, content_type_list: List[str]) -> bool:
+        """
+        Validate a list content_type values from a Datacite XML record.
+
+        Overrides the base Datacite.valid_content_types() method.
+
+        Args:
+            content_type_list: A list of content_type values.
+        """
+        if content_type_list[0] in [
```
hakbailey, line 34: This assumes that the `content_type_list` only has one value. If that ever changes in the parent class, this method will be incorrect and is likely to silently provide the wrong result. I would instead use the same strategy here as in the WHOAS version of this method, which will hedge against that future possibility.

transmogrifier/sources/dspace_dim.py
```diff
@@ -264,3 +268,13 @@ def get_source_record_id(cls, xml: Tag) -> str:
             xml: A BeautifulSoup Tag representing a single DSpace DIM XML record.
         """
         return xml.header.identifier.string.split(":")[2]
+
+    @classmethod
+    def valid_content_types(cls, content_type_list: List[str]) -> bool:
+        """
+        Validate a list content_type values from a Datacite XML record.
```
hakbailey, line 275: Copy/paste typo, and same comment as for the `Datacite` method docstring.

tests/fixtures/datacite/zenodo_records_containing_unaccepted_content_types.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<records>
```
hakbailey, line 2: I would remove all the optional fields except `resourceType` from this record since they aren't used by the test and make it harder to understand the difference between the two records.

tests/fixtures/dspace/whoas_records_containing_unaccepted_content_types.xml
```diff
@@ -0,0 +1,31 @@
...
+  <record xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <header>
+      <identifier>oai:darchive.mblwhoilibrary.org:1912/2641</identifier>
+      <datestamp>2020-01-28T19:30:01Z</datestamp>
+      <setSpec>com_1912_3</setSpec>
+      <setSpec>col_1912_534</setSpec>
+    </header>
+    <metadata>
+      <dim:dim xmlns:dim="http://www.dspace.org/xmlns/dspace/dim" xmlns:doc="http://www.lyncode.com/xoai" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.dspace.org/xmlns/dspace/dim http://www.dspace.org/schema/dim.xsd">
+        <dim:field mdschema="dc" element="title" lang="en">Time lapse movie of meiosis I in a living spermatocyte from the crane fly, Nephrotoma suturalis, viewed with polarized light microscopy</dim:field>
```
hakbailey, line 11: Same comment as for the other fixture (note: title is an optional field now).

tests/test_whoas.py
```diff
@@ -0,0 +1,31 @@
+import transmogrifier.models as timdex
+from transmogrifier.helpers import parse_xml_records
+from transmogrifier.sources.whoas import Whoas
+
+
+def test_whoas_get_content_type_returns_accepted_content_type():
```
hakbailey, line 6: For clarity of testing logic, the second test should:

1. Confirm that the length of `input_records` is 2 and the length of `output_records` is 1 to verify that we started with 2 records and 1 was skipped.
2. Checking for the `content_type` value of the returned record is fine, but checking the entire contents of the record seems unnecessary and adds confusion about what is being tested.

Might as well update the test names too, to match the actual logic of what we're testing now.

tests/test_zenodo.py
```diff
@@ -8,3 +9,38 @@ def test_zenodo_create_source_record_id_generates_correct_id():
     zenodo_record = next(output_records)
     assert zenodo_record.source_link == "https://zenodo.org/record/4291646"
     assert zenodo_record.timdex_record_id == "zenodo:4291646"
+
+
+def test_zenodo_get_content_type_returns_accepted_content_type():
```
hakbailey, line 14: Same comments as for whoas tests.

hakbailey (changes requested):
Just a few more changes needed in the tests and some suggested documentation updates.

ehanson8:
I'll tackle RDI-216 first when I circle back from other work

PR #22, opened 2022-07-21
RDI-216-default-content-type

ehanson8:
Given the more complicated `content_type` logic in `Datacite`, I had to set the default value in 2 places. I can also just set it at the top of the code block but I was going for consistent patterns between all the source classes. I'm evenly split on which is the better approach so let me know what you prefer.

Run transforms locally on the appropriate fixtures

```
pipenv run transform -i tests/fixtures/datacite/datacite_records.xml -o datacite-samples-timdex.json -s jpal
pipenv run transform -i tests/fixtures/dspace/dspace_dim_records.xml -o dspace-dim-samples-timdex.json -s whoas
pipenv run transform -i tests/fixtures/dspace/dspace_mets_records.xml -o dspace-mets-samples-timdex.json -s dspace
```

* https://mitlibraries.atlassian.net/browse/RDI-216

#### Developer

#### Code Reviewer

NO

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ update: install ## Update all Python dependencies
 
 ### Testing commands ###
 test: ## Run tests and print a coverage report
-	pipenv run coverage run --source=transmogrifier -m pytest
+	pipenv run coverage run --source=transmogrifier -m pytest -vv
 	pipenv run coverage report -m
 
 coveralls: test
--- a/tests/test_datacite.py
+++ b/tests/test_datacite.py
@@ -213,6 +213,7 @@ def test_datacite_transform_with_optional_fields_blank_transforms_correctly():
                 text="Digital object URL",
             )
         ],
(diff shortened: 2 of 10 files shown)


hakbailey (commented):
So this all technically works, but a much simpler way to do it would be to check the `content_type` field value in `transformer.transform()` and set the default value there like we do for `citation`. Two benefits to this besides fewer code changes: 1) it makes it very clear that we want that logic to be applied identically to _all_ sources, and 2) if you do it after generating the `citation`, you also don't need to change the citation generation logic.

ehanson8:
Happy to do it that way, I was incorrectly assuming from the success criteria that you wanted this handled in each source. I'll do another commit and let you know when it's done

ehanson8:
@hakbailey Pushed a new commit

hakbailey (approved):
👍

PR #27, opened 2022-09-01
TIMX-64-ead-transform-get-optional-fields

ehanson8:
This all works but it would be nice to get it more readable if possible so I welcome suggestions. A few notes:

* A lot of potential fields (particularly notes) were omitted to minimize noise and I generally chose to draw from only the `archdesc` element  given the potential for thousands of values for many records if we captured them below that level.
* I think we talked about this before, are we still cool with dumping descriptive info into `Funder.funder_name` or should a new `Funder` attribute be created? 
* I will create a ticket to refactor the date parsing code to use `@normal` and incorporate `DateRange`. I just wanted more space to examine the data while the rest of the code gets reviewed.

Run the transform locally using the `aspace-full-harvested-records-2022-08-19.xml` file in the dev S3 bucket, e.g.:

```
pipenv run transform -i <S3 URL for file above> -o aspace-timdex.json -s aspace
```

* https://mitlibraries.atlassian.net/browse/TIMX-64

#### Developer

#### Code Reviewer

NO


transmogrifier/sources/ead.py
```diff
@@ -23,13 +24,301 @@ def get_optional_fields(self, xml: Tag) -> dict:
...
+
+        # languages
+        if languages := xml.find_all("language", string=True):
+            fields["languages"] = [
+                lang_elem.string
+                for lang_elem in languages
+                if lang_elem.parent.parent.parent.name == "archdesc"
+            ] or None
+
+        # links
```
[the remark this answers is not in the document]
ehanson8: Agreed that we should omit `dao` links. However, as I look through the data, I'm not seeing a field where the collection-level link can be cleanly extracted, it's in `altformavailable` and `physloc` with other varied text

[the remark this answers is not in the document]
hakbailey: Is it not the same as the `source_link`? That should link to the page for the collection in ASpace, right?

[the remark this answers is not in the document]
hakbailey: I think there's still an open question out about whether we want to keep doing that...if there's not a spike ticket for this (whether to duplicate the `source_link` in `links`) would you mind creating one? And maybe we should not do it here for now so we have some different approaches to compare?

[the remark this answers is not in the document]
ehanson8: 👍 Created the ticket and I won't dupe it in this transform

transmogrifier/sources/ead.py
```diff
@@ -331,15 +400,24 @@ def get_main_titles(cls, xml: Tag) -> list[str]:
...
-        return [
-            titleproper_value
-            for titleproper_elem in xml.find_all("titleproper")
-            if (
-                titleproper_value := ", ".join(
-                    string for string in titleproper_elem.stripped_strings
-                )
-            )
-        ]
+        collection_description = xml.metadata.find("archdesc", level="collection")
```
ehanson8, line 403: I know this is repeated from `get_optional_fields` but I'm not sure if it's possible to pass this without significant changes to `Transformer`

transmogrifier/sources/ead.py
```diff
@@ -138,189 +43,353 @@ def get_optional_fields(self, xml: Tag) -> dict:
...
+
+    @classmethod
+    def crosswalk_type_value(cls, type_value: str) -> str:
+        """
+        Crosswalk type code to human-readable label.
+
+        Args:
+            type_value: A type value to be crosswalked.
+        """
+        type_crosswalk = {
```
ehanson8, line 341: Not sure if this rises to the level of `config.py` but  I'm happy to move it there

transmogrifier/sources/ead.py
```diff
@@ -138,189 +43,353 @@ def get_optional_fields(self, xml: Tag) -> dict:
...
+            "aat": "Art & Architecture Thesaurus",
+            "accessrestrict": "Conditions Governing Access",
+            "altformavail": "Alternative Form Available",
+            "acqinfo": "Acquisition Information",
+            "appraisal": "Appraisal",
+            "bibliography": "Bibliography",
+            "bioghist": "Biographical and Historical Note",
```
ehanson8, line 348: Misremembered this note name, I'll fix it on the next commit

transmogrifier/sources/ead.py
hakbailey: Yeah this seems fine.

transmogrifier/sources/ead.py
hakbailey: It's definition configuration. I would make it a separate json file, put it in a new config folder, and load it as a new config function.

hakbailey (commented):
Here are some general comments and then I think we can look at individual fields in future commits.

- I'm not sure I understand the value of the new fields organization. All but two of the fields we're extracting are within the collection description element, and of those two the `publication_information` field is probably better mapped to the `did.repository` element within the collection description (since the publication statement in the ead header is, again, referring to the publisher of the finding aid, not the collection).
- In addition, what are the cases where a record would not have a collection-level archdesc field? Since that's both a required element in EAD and contains almost all of the metadata we want, does it make sense to silently index a record with that significant of an error? I would suggest finding that element at the top of the method and if it doesn't exist, log an error and just return `None` for the whole method.
- Like we did for some of the other sources where we need to extract fields with a lot of attributes and sub-fields, we need to include a test for *all* the possible variations we might expect. Related to that, for any field that we expect to contain mixed content (most of them, sadly), we need to make sure the test data includes some mixed content in each of those fields.
- Having the function to generate a single string from an element's stripped string values is useful, but I think it would be much more extensible if we did the following:
  - First, make a method that returns a list of stripped strings (which we will want for some of these fields where the field value in TIMDEX is actually a list, like `notes`). Include an optional parameter in that method definition that takes in a list of elements to skip, with a default value of `None`. That way we can specify for each field which of the possible subelements we want to exclude, if any.
  - Then have the method that generates a single string call the above method to get the list first and have it just join the items in the list. Give this method an additional `separator` parameter so we can specify for any given field what the separator should be (in some cases a space may not be the best choice).
  - Now we can use one of the above methods for *every* field that contains mixed content. This will 1) make it very clear that we expect a field to contain mixed content, and 2) ensure a consistent approach to handling stripped strings across all fields.
- Finally, think very carefully through which fields need `recursive=False` in the call to `find`/`find_all`. I think in some cases here it's used when we don't want it and in some cases it isn't used when we do. Tehcnically `recursive=True` is never necessary since that's the default. If we want to pass `recursive=True` to explicitly indicate that we expect there to be useful elements further down the hierarchy, I think that's ok but we need to be consistent and probably include a comment explaining why we've done that.

Given the complexity of this transform/EAD in general, I wonder if it makes sense to do smaller commits/PRs for now. Maybe add/update the two methods described above with appropriate tests for them and then do batches of a few fields at a time? I'm open to however you want to approach it but doing it incrementally might make for both easier coding and easier code review as there's less to try and hold in the brain all at once :)

transmogrifier/sources/ead.py
ehanson8: A generic "load JSON file" function or specific function for the type crosswalk?

hakbailey: Maybe a generic `load_external_config()` function that can load from any JSON in the config folder, as I expect we'll eventually have language and possibly other config files there as well.

PR #30, opened 2022-09-21
TIMX-64-add-more-fields-to-get-optional-fields

ehanson8:
Run the transform locally using the aspace-full-harvested-records-2022-08-19.xml file in the dev S3 bucket, e.g.:

```
pipenv run transform -i <S3 URL for file above> -o aspace-timdex.json -s aspace
```

* https://mitlibraries.atlassian.net/browse/TIMX-64

#### Developer

#### Code Reviewer

NO

### DIFF
--- a/transmogrifier/sources/ead.py
+++ b/transmogrifier/sources/ead.py
@@ -1,9 +1,11 @@
 import logging
+from itertools import chain
 from typing import Generator, Optional, Union
 
 from bs4 import NavigableString, Tag
 
 import transmogrifier.models as timdex
+from transmogrifier.config import load_external_config
 from transmogrifier.sources.transformer import Transformer
 
 logger = logging.getLogger(__name__)
@@ -117,6 +119,80 @@ class Ead(Transformer):
 
         # format field not used in EAD
 
+        # holdings
+        for holding_element in chain(
+            collection_description.find_all("originalsloc", recursive=False),
+            collection_description.find_all("physloc", recursive=False),
+        ):
+            if holding_value := self.create_string_from_mixed_value(
+                holding_element, " ", ["head"]
+            ):
+                fields.setdefault("holdings", []).append(
+                    timdex.Holding(note=holding_value)
+                )
+
+        # identifiers
+        for id_element in collection_description_did.find_all("unitid"):
+            if id_value := self.create_string_from_mixed_value(
+                id_element,
+                " ",
+            ):
+                fields.setdefault("identifiers", []).append(
+                    timdex.Identifier(
+                        value=id_value,
+                    )
+                )
+
+        # languages
+        for language_element in collection_description.find_all(
+            "langmaterial", recursive=False
+        ):
+            if language_value := self.create_string_from_mixed_value(
+                language_element,
+                ", ",
+            ):
+                fields.setdefault("languages", []).append(language_value)
+
+        # links, omitted pending decision on duplicating source_link
+
+        # literary_form field not used in EAD
+
+        # locations
+        for control_access_element in collection_description.find_all(
+            "controlaccess", recursive=False
+        ):
+            for location_element in control_access_element.find_all("geogname"):
+                if location_value := self.create_string_from_mixed_value(
+                    location_element,
+                    " ",
+                ):
+                    fields.setdefault("locations", []).append(
+                        timdex.Location(value=location_value)
+                    )
+
+        # notes
+        for note_elem in [
+            note_elem
+            for note_elem in chain(
+                collection_description.find_all("acqinfo", recursive=False),
+                collection_description.find_all("appraisal", recursive=False),
+                collection_description.find_all("bibliography", recursive=False),
+                collection_description.find_all("bioghist", recursive=False),
+                collection_description.find_all("custodhist", recursive=False),
+                collection_description.find_all("processinfo", recursive=False),
+                collection_description.find_all("scopecontent", recursive=False),
+            )
+        ]:
+            if note_value := self.create_string_from_mixed_value(
+                note_elem, " ", ["head"]
+            ):
+                fields.setdefault("notes", []).append(
+                    timdex.Note(
+                        value=[note_value],
+                        kind=self.crosswalk_type_value(note_elem.name),
+                    )
+                )
+
         return fields
 
     @classmethod
@@ -163,6 +239,18 @@ class Ead(Transformer):
         return separator.join(
             cls.create_list_from_mixed_value(xml_element, skipped_elements)
         )
+
+    @classmethod
+    def crosswalk_type_value(cls, type_value: str) -> str:
+        """
+        Crosswalk type code to human-readable label.
+        Args:
+            type_value: A type value to be crosswalked.
+        """
+        type_crosswalk = load_external_config("config/aspace_type_crosswalk.json")
+        if type_value in type_crosswalk:
+            type_value = type_crosswalk[type_value]
+        return type_value
 
     @classmethod
     def generate_name_identifier_url(cls, name_element: Tag) -> Optional[list]:
--- a/transmogrifier/sources/ead.py
+++ b/transmogrifier/sources/ead.py
@@ -41,6 +41,10 @@ class Ead(Transformer):
             )
             return None
 
+        controlaccess = collection_description.find_all(
+            "controlaccess", recursive=False
+        )
+
         # alternate_titles
 
         # If the record has more than one main title, add extras to alternate_titles
@@ -63,9 +67,7 @@ class Ead(Transformer):
 
         # content_type
         fields["content_type"] = ["Archival materials"]
-        for control_access_element in collection_description.find_all(
-            "controlaccess", recursive=False
-        ):
+        for control_access_element in controlaccess:
             for content_type_element in control_access_element.find_all("genreform"):
                 if content_type_value := self.create_string_from_mixed_value(
                     content_type_element,
@@ -118,6 +120,9 @@ class Ead(Transformer):
 
         # format field not used in EAD
 
+        # funding_information field not used in EAD. titlestmt > sponsor was considered
+        # but did not fit the usage of this field in other sources.
+
         # holdings
         for holding_element in collection_description.find_all(
             "originalsloc", recursive=False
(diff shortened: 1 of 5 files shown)


transmogrifier/sources/ead.py
```diff
@@ -117,6 +119,80 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
 
         # format field not used in EAD
 
+        # holdings
```
hakbailey, line 122: Was the `funding_information` field skipped intentionally? If so, it at least needs a placeholder comment.

transmogrifier/sources/ead.py
```diff
@@ -117,6 +119,80 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
+        for holding_element in chain(
+            collection_description.find_all("originalsloc", recursive=False),
+            collection_description.find_all("physloc", recursive=False),
```
hakbailey, line 125: Per [this previous PR comment](https://github.com/MITLibraries/transmogrifier/pull/27#discussion_r963995393), `physloc` should be mapped to the `holding.location` field.

hakbailey (changes requested):
There are some general issues that need to be addressed before I can complete a review on this, plus a couple of inline comments on specific fields:
- First and most importantly, some of the new fields in the test data 1) don't match the real source data from ASpace, and 2) aren't in the correct location for valid EAD. The logic for these fields in the transform works with the test data but doesn't result in the expected output with our real harvested records.
- Given our several previous discussions about not abbreviating variable names, that needs to be double-checked and fixed where it exists in the new code.
- We've had several PR discussions about using the built-in features of bs4 instead of writing custom code to get the same functionality. Another example of that is the [ability to pass a list of elements](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#a-list) to find_all(), which would be very useful in at least one of the fields added in this PR. Since we are using bs4 in all of these transforms, I would encourage you to thoroughly read its full documentation and refer back to it anytime you need to do something new with a field extraction. It's a really neat library that has already anticipated a lot of the XML use cases we're facing.
- The new method needs unit tests for the multiple possible conditions. If you make it a staticmethod, you can write unit tests for it without creating an instance of the Ead class (which is also how we will refactor the other methods to allow for better unit testing in  future maintenance work).

I'll be on vacation tomorrow and Monday, so feel free to take all of that time to work carefully through this. This EAD transform is fiddly stuff that requires a lot of time and attention to detail to get right...but we need to get it right. Happy to answer questions if you have any this afternoon or after I get back!

transmogrifier/sources/ead.py
ehanson8: That was an oversight, I will add it on the next PR. Good catch

ehanson8:
`langmaterial` and `physloc` have been shifted to the proper locations in the test data and updated to more closely reflect the typical values in Aspace EAD

transmogrifier/sources/ead.py
ehanson8: I had previously mapped the `sponsor` element to `funding_information` but after another look and given our previous conversations about data outside of `archdesc` that relates only to the finding aid itself, I don't think `funding_information` should be mapped for EAD but I am flexible if you disagree. The values mostly refer to funding for the processing of collections which doesn't seem to fit the usage of this field in other sources.

transmogrifier/sources/ead.py
```diff
@@ -117,6 +118,86 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
+                if location_value := self.create_string_from_mixed_value(
+                    location_element,
+                    " ",
+                ):
+                    fields.setdefault("locations", []).append(
+                        timdex.Location(value=location_value)
+                    )
+
+        # notes
+        for note_element in collection_description.find_all(
```
hakbailey, line 179: I think the notes field could use more consideration:

- For some if not all of these, instead of concatenating all of the paragraphs into strings I think it makes more sense to just append each of the `<p>` tag contents to the `note.value` list. 
- `bibliography` appears to have a different structure but definitely lends itself better to a list than a single string.
- It looks like the contents of the `<head>` tag might be a better fit for the `note.kind` than the crosswalked type value for some of these.
- I'm not sure the data in `acqinfo`, `appraisal`, `custodhist`, or `processinfo` seem useful for discovery purposes? Let's review.

hakbailey (changes requested):
See inline comments, most of these are small changes but I think the notes field needs some attention. Let me know if you have questions or thoughts!

transmogrifier/sources/ead.py
```diff
@@ -117,6 +118,86 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
+        for holding_element in collection_description.find_all(
+            "originalsloc", recursive=False
+        ):
+            if holding_value := self.create_string_from_mixed_value(
+                holding_element, " ", ["head"]
+            ):
+                fields.setdefault("holdings", []).append(
+                    timdex.Holding(note=holding_value)
+                )
+        for holding_element in collection_description_did.find_all(
```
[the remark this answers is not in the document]
ehanson8: Do you think we should be creating a single `Holding` with `note=originalsloc` and `location=physloc` (if those value exist)?

Or should there be conditional block like this:

```
if collection_description.find("originalsloc"):
  ...create Holding
elif collection_description_did.find("originalsloc"):
  ...create Holding
```

Re: multiple `phyloc`'s, we've generally been treating instances of a field as separate entities, aside from when we're forced to concatenate into a string (the upcoming `physical_description`). But if it makes sense to dump them all in a string for a single `Holding` record, I can do that.

transmogrifier/sources/ead.py
```diff
@@ -117,6 +118,86 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
+            for language_value in self.create_list_from_mixed_value(language_element):
+                if language_value not in ",.":
+                    fields.setdefault("languages", []).append(language_value)
+
+        # links, omitted pending decision on duplicating source_link
+
+        # literary_form field not used in EAD
+
+        # locations
+        for control_access_element in collection_description.find_all(
```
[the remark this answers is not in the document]
ehanson8: Good call

transmogrifier/sources/ead.py
```diff
@@ -117,6 +118,86 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
+                        value=id_value,
+                    )
+                )
+
+        # languages
+        for language_element in collection_description_did.find_all(
+            "langmaterial", recursive=False
+        ):
+            for language_value in self.create_list_from_mixed_value(language_element):
+                if language_value not in ",.":
```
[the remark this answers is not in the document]
ehanson8: That is a better approach, thanks!

transmogrifier/sources/ead.py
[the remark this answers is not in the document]
hakbailey: Well, the question for this field is: conceptually, when we encounter both an `originalsloc` and a `physloc` field in a record, do they represent the same holding or different holdings? Same question with multiple `physloc` fields in a record. And if they are different holdings, how do we know and how is that represented in the data? My sense about archival collections is that there is really only one "holding" per collection, because the items are unique. That may not be true for originals vs duplicates or physical vs digital versions, but if the data doesn't clearly indicate why they are showing up as separate holdings then we may be causing confusion. This might be an area where we should discuss it further with the archives staff to better understand how they conceptualize it.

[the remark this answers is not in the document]
ehanson8: I'll get a slack chat or email going on this with us and some DDC staff to get their perspective

transmogrifier/sources/ead.py
ehanson8: 1. I'm creating a list of the `p` tags for `bioghist` and `scopecontent` but I still think we should concatenate the contents of each `p` tag given the likelihood of subelements. Let me know if you agree
2. Separating out `bibliography` from the other notes given the different structure and I'm updating the fixture for a better representation of our local data. 
3. I'm OK doing it this way and you can review how I did it but I think this adds complexity for largely insignificant changes in the `kind` value for the remaining notes `bibliography`, `bioghist`, and `scopecontent`. I scanned our data and the changes would be a few specific `bibliography` values with a person's name, `bioghist` might be `Biographical Note` or `Historical Note` rather than `Biography or History`, and the equivalent variation for `scopecontent`. It also opens up the risk of the `head` element being missing (however unlikely) and then we have a unspecified note as you can see in the updated attribute and subfield variations test. But I defer to you on whether that's a meaningful difference to users. Also, there's duplicative code here that I'll clean up after we confirm the approach
4. Looking at the data again (and as you can tell by #3), I agree!

transmogrifier/sources/ead.py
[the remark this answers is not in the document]
ehanson8: Yes, that chat happened after I submitted this commit, I already have it in place for the next commit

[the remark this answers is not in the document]
hakbailey: Is there any particular reason not to push that change as new commit for this PR? I think 1) the holdings field conversation and decision to omit it for now are part of this PR so should be included in it, and 2) I would prefer not to approve a PR that contains code we intend to immediately delete.

transmogrifier/sources/ead.py
hakbailey: 1. Yeah making the contents of each individual `<p>` tag a string makes sense.

2. Similar to the other notes with the `<p>` tag, it looks like for bibliography each element in the list should be the concatenated contents of a `<bibref>` tag. The bibliography notes are essentially a list of bibliographic references, so each of those should be an element in our `note.value` list. Right now the different parts of each bibliographic reference are getting split into separate list items, which doesn't really make sense. 

3. For the `kind` value, I don't think a few extra lines of code is significant added complexity and the `<head>` element value is more accurate, even if only slightly. We can easily fall back to the crosswalked type value if there's no `<head>` element to eliminate that problem, e.g.:
```
note_head_element = note_element.find("head", string=True)
note.kind = (
    note_head_element.string
    if note_head_element
    else self.crosswalk_type_value(note_element.name)
)
```

ehanson8: 1. 👍 
2. Done!
3. 👍 I also refactored this block to minimize duplicate code

hakbailey (approved):
Looks good!

PR #33, opened 2022-10-11
Timx 18 marc transform initial methods

ehanson8:
This is an initial approach for the `Marc` transform to revise before adding more fields. I will add inline comments for particular areas of discussion.

It has 2 commits, the first of which contains changes that affect `config`, `helpers` and `cli` as well as the necessary data for testing those changes. The 2nd commit adds methods to the `Marc` class

Run the transform locally using the `full-extracted-records-to-index-2022-09-12.xml` file in the dev S3 bucket, e.g.:

```
pipenv run transform -i <S3 URL for the above file> -o alma-transformed-records.json -s alma
```

**Note: we likely want a larger file in our dev1 bucket for more thorough manual testing going forward

#### Developer

#### Code Reviewer

NO


tests/fixtures/marc/marc_record_attribute_and_subfield_variations.xml
```diff
@@ -0,0 +1,169 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ehanson8, line 1: Let me know if this is overkill for the subfields in this fixture since there isn't a change in behavior for particular subfields

tests/test_marc.py
```diff
@@ -0,0 +1,195 @@
...
+                kind="Added Entry - Uniform Title",
+            ),
+            timdex.AlternateTitle(
+                value="Added Entry Uncontrolled Related/Analytical Title.",
+                kind="Added Entry - Uncontrolled Related/Analytical Title",
+            ),
+        ],
+        call_numbers=["MA123.4", "123.45"],
+        citation=(
+            "Célébration : 10 siècles de musique de noël.. "
```
ehanson8, line 38: MARC punctuation makes the citation a bit ugly, not sure if/how we want to address that

transmogrifier/sources/marc.py
```diff
@@ -1,8 +1,200 @@
...
+
+        Overrides metaclass get_optional_fields() method.
+
+        Args:
+            xml: A BeautifulSoup Tag representing a single MARC XML record.
+        """
+        fields: dict = {}
+
+        # alternate_titles
+        alternate_title_marc_data = {
```
ehanson8, line 26: Given the list comprehension mess I created in my initial `EAD` work, I was going for a balance of readability and abstraction with this config approach but I can see how more complex methods could reduce duplicate code in `get_optional_fields`. Let me know how you see this happening and where you'd like the config data to live

transmogrifier/sources/marc.py
```diff
@@ -1,8 +1,200 @@
...
+                    fields.setdefault("alternate_titles", []).append(
+                        timdex.AlternateTitle(
+                            value=alternate_title_value,
+                            kind=marc_field_data[1],
+                        )
+                    )
+
+        # call_numbers
+        call_numbers_marc_data = {
+            "050": ("a", ""),
```
ehanson8, line 51: I'm assuming based on `mario` that we're just grabbing the classification number but let me know if you want that to change

transmogrifier/sources/marc.py
```diff
@@ -1,8 +1,200 @@
...
+        # languages
+
+        # links
+
+        # literary_form
+        # Literary form is applicable to Book (BK) material configurations and indicated
+        # by leader "Type of Record" position = "Language Material" or "Manuscript
+        # language material" and "Bibliographic level" position =
+        # "Monographic component part," "Collection," "Subunit," or "Monograph/Item."
+        if leader := xml.find("leader", string=True):
```
ehanson8, line 93: If `leader` and `fixed_length_data` end up being called multiple times, I will move them to the top as we did with a few `Ead` variables

transmogrifier/sources/marc.py
```diff
@@ -1,8 +1,200 @@
...
+        # holdings
+
+        # identifiers
+
+        # languages
+
+        # links
+
+        # literary_form
+        # Literary form is applicable to Book (BK) material configurations and indicated
```
ehanson8, line 89: This is the kind of comment I'm planning to add whenever there is MARC logic that we have to apply

transmogrifier/sources/marc.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Create a list of values from the specified subfields of a
+        datafield element.
+
+        Args:
+            xml_element: A BeautifulSoup Tag representing a single MARC XML element.
+            subfield_codes: The codes of the subfields to extract.
+        """
+        value_list = []
+        for subfield in xml_element.children:
```
hakbailey, line 175: I think you could simplify this by just searching for what you want:
```
for subfield in xml_element.find_all(True, string=True):
    etc.
```
Then you only have to check for the right code.

tests/fixtures/marc/marc_record_attribute_and_subfield_variations.xml
hakbailey: It's probably good to include them all just for thorough testing.

transmogrifier/sources/marc.py
hakbailey: Super useful, great idea!

hakbailey (changes requested):
Generally this approach looks good, see inline comments for specific changes and comments. One general thought is that we should make sure repeatable fields are repeated in one of the test fixtures so we can confirm they're extracted as expected (including repeatable subfields within a single field, see comment on call numbers field). Probably makes the most sense in the all fields fixture, just adding one more instance of the field/subfield for anything that's repeatable?

transmogrifier/helpers.py
```diff
@@ -67,11 +67,14 @@ def generate_citation(extracted_data: dict) -> str:
     return citation
 
 
-def parse_xml_records(input_file_path: str) -> Iterator[Tag]:
+def parse_xml_records(
+    input_file_path: str,
+    xml_record_tag: str = "{http://www.openarchives.org/OAI/2.0/}record",
```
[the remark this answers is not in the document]
ehanson8: Much cleaner, thanks!

transmogrifier/sources/marc.py
ehanson8: That's a lot clearer, thanks!

transmogrifier/sources/marc.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Args:
+            xml: A BeautifulSoup Tag representing a single MARC XML record.
+        """
+        try:
+            main_title_values = []
+            if main_title_value := Marc.create_subfield_value_string_from_datafield(
+                xml.find("datafield", tag="245"), "abfgknps", " "
+            ):
+                main_title_values.append(main_title_value.rstrip(".,/"))
```
hakbailey, line 170: I think we also need to include spaces here as many records now have trailing whitespace in the title output. Probably true for alternate titles too.

hakbailey (changes requested):
There are a couple minor inline comments but the main change needed is to remove the extra code still leftover from the prior XML record parsing approach.

transmogrifier/sources/marc.py
ehanson8: Sorry, mistakenly thought whitespace was included by default, fixed!

transmogrifier/sources/marc.py
```diff
@@ -0,0 +1,225 @@
...
+            return main_title_values
+        except AttributeError:
+            logger.error(
```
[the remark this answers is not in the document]
ehanson8: I remember this now, fixed!

hakbailey (approved):
👍

PR #36, opened 2022-11-01
TIMX-18-marc-transform-more-fields

ehanson8:
Run the transform locally using the `alma-2022-10-11-full-extracted-records-to-index.xml` file in the dev S3 bucket, e.g.:

```
pipenv run transform -i <S3 URL for the above file> -o alma-transformed-records.json -s alma
```

#### Developer

#### Code Reviewer
- [x] The changes have been verified

NO

### DIFF
--- a/tests/fixtures/marc/marc_record_all_fields.xml
+++ b/tests/fixtures/marc/marc_record_all_fields.xml
@@ -56,6 +56,13 @@
             <subfield code="a">Varying Form</subfield>
             <subfield code="b">Of Title 2.</subfield>
         </datafield>
+        <datafield tag="250" ind1=" " ind2=" ">
+            <subfield code="a">9th ed. /</subfield>
+            <subfield code="b">Nick Ray ... [et al.].</subfield>
+        </datafield>
+        <datafield tag="250" ind1=" " ind2=" ">
+            <subfield code="a">Unabridged.</subfield>
+        </datafield>
         <datafield tag="264" ind1=" " ind2="1">
             <subfield code="a">France :</subfield>
             <subfield code="b">Aparte,</subfield>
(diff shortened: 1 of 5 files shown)


tests/test_marc.py
```diff
@@ -147,7 +147,72 @@ def test_marc_record_all_fields_transform_correctly():
                 kind="singer",
             ),
         ],
+        edition="9th ed. / Nick Ray ... [et al.]. Unabridged.",
         literary_form="Nonfiction",
+        subjects=[
+            timdex.Subject(
+                value=[
+                    "Renoir, Jean, 1894-1979 (DLC)n  79018203 Bibliography. "
```
ehanson8, line 155: Go was picking up the subfield 0 but I think this clutters the value and we don’t have anywhere to store it separately in the data model. With the local data, take a look at ID 990015001040106761 for an example and let me know what you think (also about the formatting in general, some values look better than others but removing subfield 0 would likely fix a lot)

JPrevost: This is interesting and I agree that we should drop $0. I suspect that when we did the original work, we weren't using $0 in practice in Aleph but we're seeing it used more now in Alma and it's indeed not appropriate for this use case. (We may want it for some fancy subject enhancement later, but for now...definitely not).

I'm seeing, in production, stuff like:
```
          "subjects": [
            "Money History 18th century. France",
            "Assignats History.",
            "Money Social aspects History 18th century. France",
            "Money Political aspects History 18th century. France",
            "Monetary policy History 18th century. France",
            "(OCoLC)fst00819080 Assignats.",
            "(OCoLC)fst00901974 Economic history.",
            "(OCoLC)fst01025230 Monetary policy.",
            "(OCoLC)fst01025265 Money.",
            "(OCoLC)fst01025293 Money Political aspects.",
            "(OCoLC)fst01025301 Money Social aspects.",
            "(OCoLC)fst01919741 Politics and government",
            "France History Revolution, 1789-1799.",
            "France Economic conditions 18th century.",
            "France Politics and government 1789-1799.",
            "(OCoLC)fst01204289 France."
          ]
```
which is kinda gross. Thanks for noticing this and yes please remove $0!

JPrevost (commented):
I had a question about the test fixture that once I understand before approval.

If you don't mind removing $0 as you suggested for subjects before we merge that would be great. If you prefer to open a ticket and come back to that we can do that instead. I think the data in $0 is likely useful for us to consider enhancing our subjects with the fast hierarchy in the future... but for now let's get it out to clean up what we present as subjects.

tests/fixtures/marc/marc_record_attribute_and_subfield_variations.xml
```diff
@@ -171,6 +183,124 @@
...
             <subfield code="t">t</subfield>
         </datafield>
+        <datafield tag="520">
+            <subfield></subfield>
+            <subfield code=""></subfield>
+            <subfield code="a">a</subfield>
+        </datafield>
+        <datafield tag="600">
+            <subfield></subfield>
+            <subfield code=""></subfield>
```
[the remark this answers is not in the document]
ehanson8: Yes, that fixture is supposed to account for all variations for attributes and subfields. Because there could conceivably be subfields with no `code` attribute or no value, we just want to make sure the transform doesn't trip up on that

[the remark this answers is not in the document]
ehanson8: @hakbailey could explain it better, I'm working off her templates for other sources

ehanson8:
Pushed a new commit to remove subfield 0 and update the fixtures and test. I'm open to a different delimiter between subfields

tests/fixtures/marc/marc_record_all_fields.xml
```diff
@@ -88,50 +88,35 @@
         <datafield tag="600" ind1="1" ind2="0">
             <subfield code="a">Renoir, Jean,</subfield>
             <subfield code="d">1894-1979</subfield>
-            <subfield code="0">(DLC)n  79018203</subfield>
```
JPrevost: I think we might want to leave these 0 and 2 subfields in the sample fixture. That way we know when they don't show up in the tests that things are working as expected.

JPrevost (approved):
I noted we may want to revert the latest changes to the fixtures to ensure we are actually confirming in our tests we are ignoring those subfields, but that's up to you if you agree and want to change or not I won't block on that opinion.

The delimiter you chose seems good to me. We can adjust in the future if we find it awkward to work with on the consuming side.

:shipit:

tests/fixtures/marc/marc_record_all_fields.xml
ehanson8: We had focused on keeping the fixtures slimmed down to the essentials so I think I want to keep it that way for now but I think this is worth discussing later on

PR #40, opened 2022-12-08
TIMX-18-dates-languages-fields

ehanson8:
Run the transform locally using the alma-2022-10-11-full-extracted-records-to-index.xml file in the dev S3 bucket, e.g.:

```
pipenv run transform -i <S3 URL for the above file> -o output/alma-transformed-records.json -s alma
```

#### Developer

#### Code Reviewer

NO

### DIFF
--- a/tests/test_marc.py
+++ b/tests/test_marc.py
@@ -507,6 +507,7 @@ def test_marc_record_attribute_and_subfield_variations_transforms_correctly():
             "k",
             "m",
             "n",
+            "Aljamía",
         ],
         links=[
             timdex.Link(url="u", kind="3", restrictions="z", text="y"),
@@ -514,7 +515,7 @@ def test_marc_record_attribute_and_subfield_variations_transforms_correctly():
         ],
         literary_form="Fiction",
         locations=[
-            timdex.Location(value="France", kind="Place of Publication"),
+            timdex.Location(value="Vietnam, North", kind="Place of Publication"),
             timdex.Location(
                 value="a - b - c - d - e - f - g - h",
                 kind="Hierarchical Place Name",
@@ -701,3 +702,23 @@ def test_get_source_record_id():
         "tests/fixtures/marc/marc_record_all_fields.xml"
     )
     assert Marc.get_source_record_id(next(marc_xml_records)) == "990027185640106761"
+
+
+def test_invalid_country_code_logs_warning(caplog):
+    marc_xml_records = parse_xml_records(
+        "tests/fixtures/marc/marc_record_attribute_and_subfield_variations.xml",
+    )
+    next(Marc("alma", marc_xml_records))
+    assert (
+        "Record # 990027185640106761 uses an obsolete location code: vn" in caplog.text
+    )
+
+
+def test_invalid_language_code_logs_warning(caplog):
+    marc_xml_records = parse_xml_records(
+        "tests/fixtures/marc/marc_record_attribute_and_subfield_variations.xml",
+    )
+    next(Marc("alma", marc_xml_records))
+    assert (
+        "Record # 990027185640106761 uses an obsolete language code: ajm" in caplog.text
+    )
--- a/tests/test_marc.py
+++ b/tests/test_marc.py
@@ -496,17 +496,15 @@ def test_marc_record_attribute_and_subfield_variations_transforms_correctly():
         ],
         languages=[
             "No linguistic content",
-            "a",
-            "b",
-            "d",
-            "e",
-            "f",
-            "g",
-            "h",
-            "j",
-            "k",
-            "m",
-            "n",
+            "English",
+            "French",
+            "Spanish",
+            "German",
+            "Russian",
+            "Chinese",
+            "Artificial (Other)",
+            "Abkhaz",
+            "Achinese",
             "Aljamía",
         ],
         links=[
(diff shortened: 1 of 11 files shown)


tests/test_marc.py
```diff
@@ -55,7 +55,8 @@ def test_marc_record_all_fields_transform_correctly():
             "Dewey Call Number 3",
         ],
         citation=(
-            "Célébration : 10 siècles de musique de noël. Language material. "
+            "Célébration : 10 siècles de musique de noël. 2016. "
```
ehanson8, line 58: I don’t think publication_information works in the citation for the Marc transform but leaving it here for your review

transmogrifier/sources/marc.py
```diff
@@ -181,6 +191,12 @@ def get_optional_fields(self, xml: Tag) -> dict:
         fields["contributors"] = contributor_values or None
 
         # dates
+        if fixed_length_data:
```
ehanson8, line 194: I kept this simple for now but I'll add handling the 2nd date in the ticket I create

transmogrifier/sources/marc.py
```diff
@@ -303,6 +385,16 @@ def get_optional_fields(self, xml: Tag) -> dict:
                 "kind": "Hierarchical Place Name",
             },
         ]
+        if fixed_length_data:
```
ehanson8, line 388: I feel like there should be a less duplicative way to handle this but I couldn't come up with anything, let me know if you see an approach

hakbailey (changes requested):
See inline comments, I think we can make the config loading more extensible.

transmogrifier/sources/marc.py
```diff
@@ -246,6 +262,72 @@ def get_optional_fields(self, xml: Tag) -> dict:
                     )
 
         # languages
+        language_marc_fields = [
+            {
+                "tag": "041",
+                "subfields": "a",
+            },
+            {
```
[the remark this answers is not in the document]
ehanson8: No good reason, this was how it was done in `mario` but I now see that's unnecessary!

tests/test_marc.py
ehanson8: Done

transmogrifier/config.py
```diff
@@ -83,10 +85,30 @@ def get_transformer(source: str) -> type:
...
+    xml_content: Tag, element_name: str, key_tag: str, value_tag: str
+) -> dict:
+    """
+    Return dict from XML object that is formatted like the country and language code
+    XML files hosted by Library of Congress.
+
+    """
+    config_dict = {}
+    for element in xml_content.find_all(element_name):
+        config_dict[element.find(key_tag).string] = element.find(value_tag).string
```
hakbailey, line 113: There's a ticket to go through the entire codebase and do this correctly, but in this case it's important now to go ahead and convert these bs4 `NavigableString` objects to Unicode strings because these config dicts will be stored in memory throughout the entire transform process. [Per the bs4 documentation](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#navigablestring), each `NavigableString` object "will carry around a reference to the entire Beautiful Soup parse tree, even when you’re done using Beautiful Soup. This is a big waste of memory."

transmogrifier/config.py
```diff
@@ -83,10 +85,30 @@ def get_transformer(source: str) -> type:
...
+        with open(file_path, "rb") as json_file:
+            return json.load(json_file)
+    elif file_type == "xml":
+        with open(file_path, "rb") as xml_file:
+            return BeautifulSoup(xml_file, "xml")
+    else:
+        return {}
+
+
+def create_dict_from_xml_config(
```
hakbailey, line 103: I would probably specify in the function name that this is for MARC or LOC XML configs.

transmogrifier/config.py
```diff
@@ -83,10 +85,30 @@ def get_transformer(source: str) -> type:
...
+    xml_content: Tag, element_name: str, key_tag: str, value_tag: str
+) -> dict:
+    """
+    Return dict from XML object that is formatted like the country and language code
+    XML files hosted by Library of Congress.
+
+    """
+    config_dict = {}
+    for element in xml_content.find_all(element_name):
```
hakbailey, line 112: Do we care about the obsolete codes that will not be found here because they're nested further down in the documents? Do we think obsolete language or country codes might exist in any of our records?

hakbailey (changes requested):
See inline for questions and comments.

transmogrifier/config.py
ehanson8: Good catch, I updated the function

transmogrifier/config.py
```diff
@@ -83,10 +87,43 @@ def get_transformer(source: str) -> type:
...
+    parameter is passed.
+
+    """
+    with open(file_path, "rb") as config_file:
+        if file_type == "json":
+            return json.load(config_file)
+        elif file_type == "xml":
+            return BeautifulSoup(config_file, "xml")
+        else:
+            logger.error("Unrecognized file_type parameter: %s", file_type)
```
hakbailey, line 104: This still isn't raising an error.

transmogrifier/config.py
```diff
@@ -83,10 +87,43 @@ def get_transformer(source: str) -> type:
...
+            label_tag: The element containing the preferred label for the item.
     """
-    with open(path, "rb") as json_file:
-        return json.load(json_file)
+    config_dict = {}
+    for code_element in xml.find_all(code_tag):
+        for parent in [
+            parent for parent in code_element.parents if parent.name == element_tag
+        ]:
+            config_dict[str(code_element.string)] = str(parent.find(label_tag).string)
```
hakbailey, line 128: We aren't testing that this code does what's expected. Also, let's think through what we want to do with obsolete codes and why. What does it mean if we encounter an obsolete code in our records? What do we want to do with that information and how do we want to handle it in the TIMDEX record?

hakbailey (approved):

PR #48, opened 2023-02-03
TIMX-180-date-cleanup

ehanson8:
Run the following commands on the specified files in the dev s3 bucket to confirm that the updated transforms still function and now log date errors. 

```
pipenv run transform -i jpal-2023-01-12-full-extracted-records-to-index.xml -o output/jpal-transformed-records.json -s jpal 

pipenv run transform -i whoas-2022-10-26-daily-extracted-records-to-index.xml -o output/whoas-transformed-records.json -s whoas 

pipenv run transform -i aspace-full-harvested-records-2022-08-19.xml -o output/aspace-transformed-records.json -s aspace

pipenv run transform -i alma-2022-10-11-full-extracted-records-to-index.xml -o output/alma-transformed-records.json -s alma 

pipenv run transform -i dspace-2022-10-18-full-extracted-records-to-index.xml -o output/dspace-transformed-records.json -s dspace
```

* https://mitlibraries.atlassian.net/browse/TIMX-180

#### Developer

#### Code Reviewer

NO


transmogrifier/helpers.py
```diff
@@ -83,6 +85,86 @@ def parse_xml_records(
...
+    Validate a date range by ensuring that the start date is before the end date to avoid
+    an OpenSearch exception. Returns true if only one date exists in the range or the end
+    date is after the start date, otherwise returns False and logs an error.
+
+    Args:
+        gte_date: The start date of a date range.
+        lte_date: The end date of a date range.
+        source_record_id: The ID of the record being transformed.
+    """
+    formatted_gte_date = format_date(gte_date)
```
hakbailey, line 151: This should have two separate checks with two separate logging statements: one to validate that both values are valid date formats, and one to confirm that the end date is equal to or greater than the start date. Although the date validation function will also log an error in the first case, it won't be obvious that the issue was in a date range as opposed to a single date unless we also log it here.

transmogrifier/helpers.py
```diff
@@ -83,6 +85,86 @@ def parse_xml_records(
...
+    formatted_lte_date = format_date(lte_date)
+    if formatted_gte_date and formatted_lte_date:
+        date_diff = formatted_gte_date - formatted_lte_date
```
hakbailey, line 154: Two things, first I would ditch the lt/gt language which is specific to OpenSearch and not super intuitive for humans. Using something like `start_date` and `end_date` here will be clearer. Also date objects can be directly compared like so: 
```
if start_date > end_date:
    we have a problem
```

hakbailey (changes requested):
See inline comments in the config and helpers modules. The source transforms will need to be updated based on those changes so I'll hold off on reviewing those for now. Let me know if you have thoughts/questions.

transmogrifier/helpers.py
```diff
@@ -83,6 +85,90 @@ def parse_xml_records(
...
+def parse_date_from_string(
+    date_string: str,
+) -> Optional[datetime]:
+    """
+    Transform a date string into a datetime object according to one of the configured
+    OpenSearch date formats.
+
+    Args:
+        date_string: A date string.
+        source_record_id: The ID of the record being transformed.
```
hakbailey, line 97: This arg should be removed from the docstring. I would probably also note here that this returns `None` if the date string can't be parsed.

transmogrifier/helpers.py
```diff
@@ -83,6 +85,90 @@ def parse_xml_records(
...
+    Validate that a date string can be parsed according to one of the configured
+    OpenSearch date formats. Returns True if date string is valid or returns
+    False and logs an error.
+
+    Args:
+        date_string: A date string.
+        source_record_id: The ID of the record being transformed.
+    """
+    date_string = date_string.strip()
+    try:
```
hakbailey, line 121: I'm not quite sure why this is in a try/catch block. `parse_date_from_string()` shouldn't raise a `ValueError`, it will either return a date or `None` right?

tests/test_helpers.py
```diff
@@ -184,3 +192,101 @@ def test_generate_citation_with_all_fields():
...
     assert len(list(records)) == 38
+
+
+def test_parse_date_from_string_success():
+    for date in [
+        "1930",
+        "1930-12",
+        "1930",
+        "1930-12",
+        "1930-12-31",
```
hakbailey, line 203: Is there a reason for the duplicate date strings in here?

tests/test_helpers.py
```diff
@@ -184,3 +192,101 @@ def test_generate_citation_with_all_fields():
...
+        "1930-12-31T12:34Z",
+        "1930-12-31T12:34:56",
+        "1930-12-31T12:34:56Z",
+        "1930-12-31T12:34:56.000001",
+        "1930-12-31T12:34:56.000001Z",
+    ]:
+        assert type(parse_date_from_string(date)) == datetime
+
+
+def test_parse_date_from_string_invalid_date_returns_false():
```
hakbailey, line 262: Should probably rename this and explicitly check for `is None`.

tests/test_helpers.py
```diff
@@ -184,3 +192,101 @@ def test_generate_citation_with_all_fields():
...
+    assert not parse_date_from_string("circa 1930s")
+
+
+def test_validate_date_success():
+    assert validate_date("1930 ", "1234") is True
+
+
+def test_validate_date_invalid_date_logs_error(caplog):
```
hakbailey, line 270: This should also confirm it returns `False`. Same for date range validation tests below.

transmogrifier/sources/datacite.py
```diff
@@ -117,15 +118,39 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
         for date in xml.metadata.find_all("date"):
             d = timdex.Date()
-            if value := date.string:
-                if "/" in value:
-                    split = value.index("/")
-                    d.range = timdex.Date_Range(
-                        gte=value[:split],
-                        lte=value[split + 1 :],
-                    )
+            if date_value := date.string:
```
hakbailey, line 121: Rather than continuing to use this check, let's update the `find_all` function above to include `string=True`.

transmogrifier/sources/datacite.py
```diff
@@ -117,15 +118,39 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
...
+                    gte_date = date_value[:split]
+                    lte_date = date_value[split + 1 :]
+                    if validate_date_range(
+                        gte_date,
+                        lte_date,
+                        source_record_id,
+                    ):
+                        d.range = timdex.Date_Range(
+                            gte=gte_date
+                            if validate_date(
```
hakbailey, line 133: You don't need these additional checks for `validate_date` since `validate_date_range` does that already. Same in the other sources.

transmogrifier/helpers.py
```diff
@@ -83,6 +85,90 @@ def parse_xml_records(
...
+
+
+def validate_date_range(
+    start_date: str,
+    end_date: str,
+    source_record_id: str,
+) -> bool:
+    """
+    Validate a date range by ensuring that the start date is before the end date to avoid
+    an OpenSearch exception. Returns true if only one date exists in the range or the end
```
hakbailey, line 141: I would maybe clarify here that this also validates that both the start and end dates can be parsed.

hakbailey (changes requested):
See inline comments, just a few more small changes.

tests/test_helpers.py
ehanson8: A find-and-replace went sideways, fixing it now

ehanson8: A find-and-replace went sideways, fixing it now

transmogrifier/sources/datacite.py
hakbailey: Ah got it, yeah that makes sense.

tests/test_helpers.py
```diff
@@ -199,38 +199,38 @@ def test_parse_date_from_string_success():
         "1930",
         "1930-12",
         "1930",
-        "1930-12",
-        "1930-12-31",
-        "1930-12-31",
-        "19301231",
+        "30-12",
```
hakbailey, line 202: Still duplicates?

hakbailey (approved):
Other than removing the remaining duplicate values in the test_helpers dates, this looks good.

PR #54, opened 2023-03-07
Fix recursion bug

hakbailey:
#### What does this PR do?
Replaces recursive call in the Transformer class `__next__` method with an internal loop.

The Transformer's internal `__next__` method used recursion to skip over records that couldn't be transformed, which was both inefficient and hit a recursion depth limit while processing deletes (none of which are transformed).

On the current main branch, try to transform the Alma file that was causing issues and you should hit the same error we did on prod (I copied the file to Dev1 for easier testing):
```
 pipenv run transform -i s3://timdex-extract-dev-222053980223/alma/alma-2023-03-07-daily-extracted-records-to-delete.xml -o output/alma-deletes.json -s alma
```
Then checkout this branch and run the same transform. It should complete successfully.

NO

#### Developer

#### Code Reviewer
- [ ] The commit message is clear and follows our guidelines (not just this pull request message)
- [ ] Why these changes are being introduced:

### DIFF
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -40,19 +40,20 @@ class Transformer(object):
 
     def __next__(self) -> TimdexRecord:
         """Return next transformed record."""
-        xml = next(self.input_records)
-        self.processed_record_count += 1
-        try:
-            record = self.transform(xml)
-        except DeletedRecord as error:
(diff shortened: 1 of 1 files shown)


ehanson8 (approved):
👍 Good write-up of the error and works as expected on dev!

PR #56, opened 2023-03-10
date-whitespace-bug-fix

ehanson8:
Though `Marc` is the transform most likely to encounter this issue, I updated the date processing for all of the transforms. Let me know if this overkill.

I uploaded a file with a whitespace date to dev1. Run the following command on `main` to see that the whitespace is in the output and then with `date-whitespace-bug-fix` to see that the date appears without whitespace:

```
pipenv run transform -i s3://timdex-extract-dev-222053980223/alma/date-error.xml -o output/date-error-transformed-records.json -s alma
```

#### Developer

#### Code Reviewer

NO

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -264,14 +264,19 @@ def test_parse_date_from_string_invalid_date_returns_none():
 
 
 def test_validate_date_success():
-    assert validate_date("1930 ", "1234") is True
+    assert validate_date("1930", "1234") is True
 
 
 def test_validate_date_invalid_date_logs_error(caplog):
     assert validate_date("circa 1930s", "1234") is False
     assert (
-        "Record # '1234' has a date that couldn't be parsed: circa 1930s"
+        "Record # '1234' has a date that couldn't be parsed: 'circa 1930s'"
     ) in caplog.text
+
+
+def test_validate_date_whitespace_date_logs_error(caplog):
+    assert validate_date("30  ", "1234") is False
+    assert ("Record # '1234' has a date that couldn't be parsed: '30  '") in caplog.text
 
 
 def test_validate_date_range_success():
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -198,7 +198,6 @@ def test_parse_date_from_string_success():
     for date in [
         "1930",
         "1930-12",
-        "30",
         "30-12",
         "30-12-31",
         "1930-12-31",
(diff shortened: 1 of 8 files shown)


tests/test_helpers.py
```diff
@@ -264,16 +264,21 @@ def test_parse_date_from_string_invalid_date_returns_none():
...
     assert validate_date("circa 1930s", "1234") is False
     assert (
-        "Record # '1234' has a date that couldn't be parsed: circa 1930s"
+        "Record # '1234' has a date that couldn't be parsed: 'circa 1930s'"
     ) in caplog.text
 
 
+def test_validate_date_whitespace_date_logs_error(caplog):
+    assert validate_date("30  ", "1234") is False
+    assert ("Record # '1234' has a date that couldn't be parsed: '30  '") in caplog.text
```
JPrevost, line 279: Not directly related to this line, but are we sure OpenSearch is treating "30" as a date the same way it might be meant in MARC? Adding the strip() prevents the problem... but are we ending up with data that matches the intent of the cataloger? I'm not saying we aren't, but since there is no test here to clarify these types of edge cases and whether we want "30" to be valid I'm not sure if we are doing the correct thing by just stripping and not confirming the intent remains for these yet.

JPrevost: Do you have a record id we can compare in OpenSearch with the record in Alma to ensure the two systems are treating two digit dates the same way?

hakbailey (changes requested):
Any time you are using a `NavigableString` as a value, you need to convert it to a regular Python string [per the bs4 docs](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#navigablestring). There's a ticket to make that change across the repo, but you should go ahead and update any code you're already touching, including all of these dates.

transmogrifier/sources/datacite.py
```diff
@@ -118,11 +120,11 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
 
         for date in xml.metadata.find_all("date"):
             d = timdex.Date()
-            if date_value := date.string:
+            if date_value := str(date.string):
```
hakbailey, line 123: `str(None)` results in `"None"`, which means this will never be False.

transmogrifier/sources/datacite.py
```diff
@@ -108,7 +108,9 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
         # dates
         if publication_year := xml.metadata.find("publicationYear", string=True):
             fields["dates"] = [
-                timdex.Date(kind="Publication date", value=publication_year.string)
+                timdex.Date(
+                    kind="Publication date", value=publication_year.string.strip()
```
hakbailey, line 112: This is also a NavigableString.

transmogrifier/sources/dspace_dim.py
```diff
@@ -88,11 +88,15 @@ def get_optional_fields(self, xml: Tag) -> Optional[dict]:
 
         # dates
         for date in [d for d in xml.find_all("dim:field", element="date") if d.string]:
```
hakbailey, line 90: While we're in here, let's simplify this.
```
for date in xml.find_all("dim:field", element="date", string=True):
```

hakbailey (changes requested):
See inline comments.

hakbailey (approved):
👍

PR #59, opened 2023-03-15
TIMX-212-invalid-date-fix

ehanson8:
After reviewing all the transforms, `Ead` appears to be the only one that was affected by this bug.

Before pulling down the new branch, run the following command with `main` while authenticated to dev:

```
pipenv run transform -i s3://timdex-extract-dev-222053980223/aspace/aspace-2023-10-17-full-extracted-records-to-index.xml -o output/aspace-records.json -s aspace
```
Check record `aspace:repositories-2-resources-531` and see the invalid date. 

Fetch `TIMX-212-invalid-date-fix` and run the same command to see that the invalid date has been omitted.

* https://mitlibraries.atlassian.net/browse/TIMX-212

#### Developer

#### Code Reviewer

NO

### DIFF
--- a/tests/fixtures/ead/ead_record_attribute_and_subfield_variations.xml
+++ b/tests/fixtures/ead/ead_record_attribute_and_subfield_variations.xml
@@ -38,15 +38,21 @@
                         <unitdate datechar="creation"></unitdate>
                         <unitdate datechar="creation">1905-2012</unitdate>
                         <unitdate datechar="creation">1905</unitdate>
+                        <unitdate datechar="creation">abcd-efgh</unitdate>
+                        <unitdate datechar="creation">abcd</unitdate>
                         <unitdate datechar="creation" certainty=""></unitdate>
                         <unitdate datechar="creation" certainty="">1905-2012</unitdate>
                         <unitdate datechar="creation" certainty="">1905</unitdate>
                         <unitdate datechar="creation" certainty="approximate"></unitdate>
                         <unitdate datechar="creation" certainty="approximate">1905-2012</unitdate>
                         <unitdate datechar="creation" certainty="approximate">1905</unitdate>
+                        <unitdate datechar="creation" certainty="approximate">abcd-efgh</unitdate>
+                        <unitdate datechar="creation" certainty="approximate">abcd</unitdate>
                         <unitdate certainty="approximate"></unitdate>
                         <unitdate certainty="approximate">1905-2012</unitdate>
                         <unitdate certainty="approximate">1905</unitdate>
+                        <unitdate certainty="approximate">abcd-efgh</unitdate>
+                        <unitdate certainty="approximate">abcd</unitdate>
                         <unitdate datechar="" certainty="approximate"></unitdate>
                         <unitdate datechar="" certainty="approximate">1905-2012</unitdate>
                         <unitdate datechar="" certainty="approximate">1905</unitdate>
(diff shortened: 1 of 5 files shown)


tests/fixtures/ead/ead_record_attribute_and_subfield_variations.xml
```diff
@@ -38,15 +38,21 @@
                         <unitdate datechar="creation"></unitdate>
                         <unitdate datechar="creation">1905-2012</unitdate>
                         <unitdate datechar="creation">1905</unitdate>
+                        <unitdate datechar="creation">abcd-efgh</unitdate>
```
JPrevost, line 41: The test suite for this app is very confusing. I'm not asking for a specific change as part of this code review as I have confirmed that reverting the code in `ead.py` to the version without the conditional check fails the suite as expected. I'm sharing this input mostly as a bit of a warning that the process of testing may not be intuitive enough for others to assist in maintaining this application. There was no additional regression tests added to the suite to ensure the block of code works as intended, yet somehow adding additional lines to a single fixture file is allowing us ensure the code logic is correct.

For unit testing, I'd expect we would have added an additional test for the new regression such as:
`test dates should only be valid if there is a value or a range`

Making a test suite that is explicit like that allows us to read through the tests to understand functionality. Again, I'm not stating this test suite doesn't test the code as I believe it does. I'm just nervous it is a bit too opaque as to how the code is supposed to work even if the tests are passing "correctly".

JPrevost: When I revert the code in ead.py, I get an error in this test (as expected):
`test_ead_record_with_attribute_and_subfield_variations_transforms_correctly`

It isn't clear _why_ the transform isn't working and I believe it will be very much harder to support this application with only "top level" testing rather than more granular tests... particularly when guarding against regressions.

ehanson8: We discussed this before and I do agree. It would be a significant overhaul of the testing suite but I agree that it's worthwhile for the long-term maintainability. I will create a ticket to do that

JPrevost: I agree that a full overhaul would take some work, but there isn't anything really preventing us from adding one or two additional tests when we introducing changes like these. The giant mega test AND small focused tests is genuinely the best of both worlds.

ehanson8: Good call, I'll add the tests in a new commit

JPrevost (approved):
Your code works and it has tests.

Let's merge this now, and if you are interested, discuss strategies to make tests a bit more readable on zoom at some point.

tests/test_ead.py
```diff
@@ -504,6 +504,28 @@ def test_ead_record_with_blank_optional_fields_transforms_correctly():
     )
 
 
+def test_ead_record_invalid_date_and_date_range_are_omitted(caplog):
+    ead_xml_records = parse_xml_records(
+        "tests/fixtures/ead/ead_record_attribute_and_subfield_variations.xml"
+    )
+    output_record = next(Ead("aspace", ead_xml_records))
+    assert "abcd" not in [d.value for d in output_record.dates]
```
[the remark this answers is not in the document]
ehanson8: 👍 Sounds good, I've created a [ticket](https://mitlibraries.atlassian.net/browse/TIMX-214) for the discussion

PR #101, opened 2023-08-04
Timx 232 springshare ids

ghukill:
#### What does this PR do?

This PR adds two new OPTIONAL methods to the base transform:
  * `get_source_link()`
  * `get_timdex_record_id()`

For transformations that require it, this allows for decoupling the logic of constructing the timdex record id from the source link.  This directly supports these new sources, and it's believed will also be helpful for future sources where this tight coupling may not be ideal.

Dependencies are also updated, with slight type check updates for new `flake8` requirements in `ead` transformation.

While analyzing the transformed records from new `libguides` and `researchdatabases` sources, it was found that some timdex record ids were poorly formed, e.g. `libguides:c.php?g=175847`.

The original approach was splitting the `dc:identifier` and using that for the timdex record id and concatenating with the source base url.  However, in a small subset, this resulted in poorly formed ids like the one above.

Given the terse nature of these source records, there is not a meaningful identifier in the record that could serve both the timdex record id _and_ the source link.

For the following fixture (showing subset of fields here):
```xml
<?xml version="1.0" encoding="UTF-8"?>
<records>
  <record>
    <header>
      <identifier>oai:libguides.com:guides/175846</identifier>  <!-- OAI identifier -->
      ...
    </header>
    <metadata>
      <oai_dc:dc>
        ...
        <dc:identifier>https://libguides.mit.edu/materials</dc:identifier>  <!-- DC identifier -->
        ...
      </oai_dc:dc>
    </metadata>
  </record>
</records>
```

Run a `libguides` transformation:
```bash
pipenv run transform -s libguides -i "tests/fixtures/oai_dc/springshare/libguides/libguides_record_all_fields.xml" -o "output/libguides.json"
```

Note the following fields are set:
```python
timdex_record_id="libguides:guides-175846"
source_link="https://libguides.mit.edu/materials",
```

Demonstrating the source link, due to the overriden method, is decoupled from the timdex id.

https://mitlibraries.atlassian.net/browse/TIMX-232

#### Developer

#### Code Reviewer

YES

### DIFF
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -118,10 +118,12 @@ class Transformer(object):
         """
         source_record_id = self.get_source_record_id(xml)
         title = self.get_valid_title(source_record_id, xml)
+        source_link = self.get_source_link(self.source_base_url, source_record_id, xml)
+        timdex_record_id = self.get_timdex_record_id(self.source, source_record_id, xml)
         return {
             "source": self.source_name,
-            "source_link": self.source_base_url + source_record_id,
-            "timdex_record_id": f"{self.source}:{source_record_id.replace('/', '-')}",
+            "source_link": source_link,
+            "timdex_record_id": timdex_record_id,
             "title": title,
         }
 
@@ -180,7 +182,7 @@ class Transformer(object):
                 source_record_id,
                 all_titles,
             )
-        if all_titles and type(all_titles[0]) == str:
+        if all_titles and isinstance(all_titles[0], str):
             title = all_titles[0]
         elif all_titles and all_titles[0].string:
             title = all_titles[0].string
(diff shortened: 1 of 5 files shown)


transmogrifier/sources/transformer.py
```diff
@@ -191,3 +193,39 @@ def get_valid_title(cls, source_record_id: str, xml: Tag) -> str:
             )
             title = "Title not provided"
         return title
+
+    @classmethod
+    def get_source_link(
+        cls, source_base_url: str, source_record_id: str, xml: Tag
```
ehanson8, line 199: Do we need `xml` as a parameter here if it's not being used? Couldn't the method just be overridden in the source transforms with an `xml` parameter when it's needed?

transmogrifier/sources/transformer.py
```diff
@@ -118,10 +118,12 @@ def get_required_fields(self, xml: Tag) -> dict:
         """
         source_record_id = self.get_source_record_id(xml)
         title = self.get_valid_title(source_record_id, xml)
+        source_link = self.get_source_link(self.source_base_url, source_record_id, xml)
+        timdex_record_id = self.get_timdex_record_id(self.source, source_record_id, xml)
```
ehanson8, line 122: Optional: you could call these directly in the dict

ehanson8 (commented):
Two comments

transmogrifier/sources/transformer.py
ghukill: Happy to change if you'd like, and I considered that too.  But at the risk of a being a bit more verbose, I thought it helped break it up a bit.  There is some important stuff happening here -- establishing record title, primary timdex id, link, etc. -- and I found it helpful to see those steps as distinct statements vs embedded in a dictionary.

transmogrifier/sources/transformer.py
ghukill: Nice catch, and I had also paused on this.  I think if these are getting decoupled with the understanding transformers may override them, it's plausible they might need the document XML, so I included here.

Because the base transform calls this method, it will always use the same signature, so a sub-class transformer couldn't add `xml:Tag` on their own.  

We could adopt a `kwargs` approach, where the sub-class transformer _could_ pass anything they want as named arguments, but we'd lose the typing this provides.  Other than `xml:Tag`, it's hard to guess what they may want to pass?  

See pros and cons both ways.  My thinking was to model these after other methods that are designed to overridden, like `get_valid_title()` or `get_source_record_id()` which both pass `xml:Tag`.

ghukill: But if helpful, we could add the docstring comment:

```
xml: A BeautifulSoup Tag representing a single XML record.
                - not used by default implementation, but could be useful for subclass
                overrides
```

which `get_timdex_record_id()` contains below.  I had actually intended to do so.

transmogrifier/sources/transformer.py
ehanson8: Optional is always just a suggestion 🙂

transmogrifier/sources/transformer.py
ehanson8: Let's just add the comment, I did some more reading and see your point on losing the type hinting

ehanson8 (approved):

PR #106, opened 2023-11-21
Gdt 82 transformer class refactor

ehanson8:
The addition of the `Transformer` base class and associated refactoring will be followed by a new `JsonTransformer` format class that will be used for upcoming `aardvark` transforms.

Run a transform on a test fixture to ensure that the base class and associated refactor hasn't broken the application

```
pipenv run transform -i tests/fixtures/ead/ead_record_all_fields.xml -o output/aspace-records.json -s aspace
```

* https://mitlibraries.atlassian.net/browse/GDT-82

#### Developer

#### Code Reviewer

YES

### DIFF
--- a/tests/test_datacite.py
+++ b/tests/test_datacite.py
@@ -1,4 +1,3 @@
-from transmogrifier.helpers import parse_xml_records
 from transmogrifier.models import (
     AlternateTitle,
     Contributor,
@@ -170,7 +169,7 @@ def test_datacite_transform_with_all_fields_transforms_correctly(
 
 
 def test_datacite_transform_missing_required_datacite_fields_logs_warning(caplog):
-    source_records = parse_xml_records(
+    source_records = Datacite.parse_source_file(
         "tests/fixtures/datacite/datacite_record_missing_datacite_required_fields.xml"
     )
     output_records = Datacite("cool-repo", source_records)
@@ -195,7 +194,7 @@ def test_datacite_transform_missing_required_datacite_fields_logs_warning(caplog
 
 
 def test_datacite_transform_with_optional_fields_blank_transforms_correctly():
-    source_records = parse_xml_records(
+    source_records = Datacite.parse_source_file(
         "tests/fixtures/datacite/datacite_record_optional_fields_blank.xml"
     )
     output_records = Datacite("cool-repo", source_records)
@@ -218,7 +217,7 @@ def test_datacite_transform_with_optional_fields_blank_transforms_correctly():
 
 
 def test_datacite_transform_with_optional_fields_missing_transforms_correctly():
-    source_records = parse_xml_records(
+    source_records = Datacite.parse_source_file(
         "tests/fixtures/datacite/datacite_record_optional_fields_missing.xml"
     )
     output_records = Datacite("cool-repo", source_records)
@@ -241,7 +240,7 @@ def test_datacite_transform_with_optional_fields_missing_transforms_correctly():
 
 
 def test_datacite_with_attribute_and_subfield_variations_transforms_correctly():
-    source_records = parse_xml_records(
+    source_records = Datacite.parse_source_file(
         "tests/fixtures/datacite/datacite_record_attribute_and_subfield_variations.xml"
     )
     output_records = Datacite("cool-repo", source_records)
--- a/tests/test_datacite.py
+++ b/tests/test_datacite.py
@@ -170,10 +170,10 @@ def test_datacite_transform_with_all_fields_transforms_correctly(
 
 
 def test_datacite_transform_missing_required_datacite_fields_logs_warning(caplog):
-    input_records = parse_xml_records(
+    source_records = parse_xml_records(
         "tests/fixtures/datacite/datacite_record_missing_datacite_required_fields.xml"
     )
-    output_records = Datacite("cool-repo", input_records)
+    output_records = Datacite("cool-repo", source_records)
     next(output_records)
 
     assert (
@@ -195,10 +195,10 @@ def test_datacite_transform_missing_required_datacite_fields_logs_warning(caplog
 
 
 def test_datacite_transform_with_optional_fields_blank_transforms_correctly():
-    input_records = parse_xml_records(
+    source_records = parse_xml_records(
         "tests/fixtures/datacite/datacite_record_optional_fields_blank.xml"
     )
-    output_records = Datacite("cool-repo", input_records)
+    output_records = Datacite("cool-repo", source_records)
     assert next(output_records) == TimdexRecord(
         citation=("Title not provided. https://example.com/doi:10.7910/DVN/19PPE7"),
         source="A Cool Repository",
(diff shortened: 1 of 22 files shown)


transmogrifier/sources/transformer.py
```diff
@@ -55,73 +68,246 @@ def __next__(self) -> TimdexRecord:
                 self.skipped_record_count += 1
                 continue
 
+    @final
+    @staticmethod
+    def get_transformer(source: str) -> Transformer:
```
ghukill, line 73: Apologies, think I misspoke in our chat, think this return type should be a `type[Transformer]` to indicate it's not yet initialized, e.g.:

```python
def get_transformer(source: str) -> type[Transformer]:
...
```

This resolves an IDE linting error in `cli.main` where an instance is not callable.

transmogrifier/sources/transformer.py
```diff
@@ -131,25 +317,25 @@ def get_required_fields(self, xml: Tag) -> dict:
         }
 
     @final
-    def transform(self, xml: Tag) -> Optional[TimdexRecord]:
+    def transform(self, source_record: Tag) -> Optional[TimdexRecord]:
         """
         Transform an OAI-PMH XML record into a TIMDEX record.
```
ghukill, line 322: Feels like an opportunity to remove "OAI-PMH" where it pops up and be generic that it's just "XML".

transmogrifier/sources/transformer.py
```diff
@@ -55,73 +68,246 @@ def __next__(self) -> TimdexRecord:
...
+    """Base transformer class."""
+
+    @final
+    @classmethod
+    def parse_source_file(cls, input_file: str) -> Iterator[Tag]:
+        """
+        Parse source file and return source records via an iterator.
+
+        May not be overridden.
+        """
```
ghukill, line 227: I would suggest updating this docstring to be XML specific, thereby differentiating from the base `Transformer`.

tests/test_datacite.py
```diff
@@ -170,10 +169,10 @@ def test_datacite_transform_with_all_fields_transforms_correctly(
 
 
 def test_datacite_transform_missing_required_datacite_fields_logs_warning(caplog):
-    input_records = parse_xml_records(
+    source_records = Datacite.parse_source_file(
```
ghukill, line 172: Just a note: as we discussed, if the getting of corrrect transformer class and parsing of input records is combined, I think a lot of these explicit calls to specific transformers could be replaced with something like:

```python
transformer = Transformer.load("datacite", "file_path.xml")
```

But, also acknowledging this is likely for a future PR.

transmogrifier/helpers.py
```diff
@@ -73,22 +67,6 @@ def generate_citation(extracted_data: dict) -> str:
...
-            tag="{*}record",
-            encoding="utf-8",
-            recover=True,
-        ):
-            record_string = etree.tostring(element, encoding="utf-8")
-            record = BeautifulSoup(record_string, "xml")
-            yield record
-            element.clear()
-
-
```
ghukill, line 91: +1 for deleted (albiet it moved...) code!

transmogrifier/cli.py
```diff
@@ -49,9 +44,9 @@ def main(source, input_file, output_file, verbose):
     logger.info(configure_sentry())
     logger.info("Running transform for source %s", source)
 
-    input_records = parse_xml_records(input_file)
-    transformer_class = get_transformer(source)
-    transformer_instance = transformer_class(source, input_records)
+    transformer_class = Transformer.get_transformer(source)
+    source_records = transformer_class.parse_source_file(input_file)
+    transformer_instance = transformer_class(source, source_records)
```
ghukill, line 49: Perhaps a baby step towards refactoring this could be renaming this variable to just `transformer`, where it's _implied_ this is a `Transformer` instance.

transmogrifier/sources/transformer.py
```diff
@@ -55,73 +68,246 @@ def __next__(self) -> TimdexRecord:
...
         Args:
-            xml: A BeautifulSoup Tag representing a single XML record
+            source_record: A BeautifulSoup Tag representing a single XML record
         """
         return {}
 
     @classmethod
     @abstractmethod
-    def get_main_titles(cls, xml: Tag) -> list[Tag]:
+    def get_main_titles(cls, source_record: Tag) -> list[Tag]:
```
ghukill, line 254: Arbitrary where to note this, but like the change from `xml` (which is problematic for shadowing a stdlib name and also just kind of generic) to `source_record`.  This will make context switching from XML to JSON easier when looking at multiple transformations.

ghukill (changes requested):
Overall, I think this looks fantastic.

The requested changes are mostly quite small, just some rough edges from a fairly big refactoring process.  

And, acknowledging per conversations that some more changes may be coming related to getting the appropriate transformer class type and parsing input records.

tests/test_datacite.py
ghukill: It's a subtle thing, but if we decide to move more functionality into the `Transformer` class, we might want to not refer to an instantiated `Transformer` as `output_records`.

Then again, if we keep as a true iterator in essence, I can the logic behind that variable naming.

transmogrifier/helpers.py
```diff
@@ -177,7 +155,7 @@ def write_deleted_records_to_file(deleted_records: list[str], output_file_path:
 
 
 def write_timdex_records_to_json(
-    transformer_instance: "Transformer", output_file_path: str
+    transformer_instance: "XmlTransformer", output_file_path: str
```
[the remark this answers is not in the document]
ehanson8: Inadvertent find-replace, restored `Transformer`

transmogrifier/sources/transformer.py
ehanson8: Missed this one!

transmogrifier/cli.py
ehanson8: Good call!

ghukill (approved):
Approved!

I did have one question, but perhaps this is slated for a future PR.  Any reason to not move `helpers.write_deleted_records_to_file()` ([link](https://github.com/MITLibraries/transmogrifier/blob/gdt-82-transformer-class-refactor/transmogrifier/cli.py#L48-L52)) to a `Transformer` method as well?

If so, I think it calls into question the nature of the method refactored method `Transformer.write_timdex_records_to_json()`.  

I wonder if a higher level method like `Transformer.write_output_files()` could encapsulate the logic and processing of both?  It could:
  * write all normal records to `output_file`
  * if deleted records, rename `output_file` and write
  * log all information that was formerly in CLI

Still marking as approved, as I think this PR represents a big and good change, but just floating that idea for future PRs.

ehanson8:
> * write all normal records to `output_file`
> * if deleted records, rename `output_file` and write
> * log all information that was formerly in CLI


Exactly what I was thinking, saving for a future PR since this was already so big!

PR #107, opened 2023-12-01
Gdt 83 jsontransformer

ehanson8:
Additional refactoring was needed on the `XmlTransformer` class to provide a proper example for the `JsonTransformer` class. Given the incomplete nature of `JsonTransformer` which has several `abstractmethod`s that need to be defined by source transforms, proper testing won't be possible until a source transform is derived from `JsonTransformer`

Run a transform on a test fixture to ensure that the base class and associated refactor hasn't broken the application

```
pipenv run transform -i tests/fixtures/ead/ead_record_all_fields.xml -o output/aspace-records.json -s aspace
```

* https://mitlibraries.atlassian.net/browse/GDT-83

#### Developer

#### Code Reviewer

NO

### DIFF
--- a/transmogrifier/cli.py
+++ b/transmogrifier/cli.py
@@ -5,7 +5,6 @@ from time import perf_counter
 import click
 
 from transmogrifier.config import SOURCES, configure_logger, configure_sentry
-from transmogrifier.helpers import write_deleted_records_to_file
 from transmogrifier.sources.transformer import Transformer
 
 logger = logging.getLogger(__name__)
@@ -42,14 +41,7 @@ def main(source, input_file, output_file, verbose):
     logger.info("Running transform for source %s", source)
 
     transformer = Transformer.load(source, input_file)
-    transformer.write_timdex_records_to_json(output_file)
-    if transformer.processed_record_count == 0:
-        raise ValueError("No records processed from input file, needs investigation")
-    if deleted_records := transformer.deleted_records:
-        deleted_output_file = output_file.replace("index", "delete").replace(
-            "json", "txt"
-        )
-        write_deleted_records_to_file(deleted_records, deleted_output_file)
+    transformer._write_output_files(output_file)
     logger.info(
         (
             "Completed transform, total records processed: %d, "
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -70,7 +70,20 @@ class Transformer(ABC):
                 continue
 
     @final
-    def write_timdex_records_to_json(self, output_file: str) -> int:
+    def _write_output_files(self, output_file: str) -> None:
+        self.write_timdex_records_to_json_file(output_file)
+        if self.processed_record_count == 0:
+            raise ValueError(
+                "No records processed from input file, needs investigation"
+            )
+        if deleted_records := self.deleted_records:
+            deleted_output_file = output_file.replace("index", "delete").replace(
+                "json", "txt"
+            )
+            self.write_deleted_records_to_txt_file(deleted_records, deleted_output_file)
+
+    @final
+    def write_timdex_records_to_json_file(self, output_file: str) -> int:
         """
         Write TIMDEX records to JSON file.
 
@@ -105,6 +118,15 @@ class Transformer(ABC):
                 file.write(",\n")
             file.write("\n]")
         return count
+
+    @final
+    @staticmethod
+    def write_deleted_records_to_txt_file(
+        deleted_records: list[str], output_file_path: str
+    ):
+        with open(output_file_path, "w") as file:
+            for record_id in deleted_records:
+                file.write(f"{record_id}\n")
 
     @final
     @classmethod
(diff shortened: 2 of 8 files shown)


transmogrifier/cli.py
```diff
@@ -42,14 +41,7 @@ def main(source, input_file, output_file, verbose):
...
     transformer = Transformer.load(source, input_file)
-    transformer.write_timdex_records_to_json(output_file)
-    if transformer.processed_record_count == 0:
-        raise ValueError("No records processed from input file, needs investigation")
-    if deleted_records := transformer.deleted_records:
-        deleted_output_file = output_file.replace("index", "delete").replace(
-            "json", "txt"
-        )
-        write_deleted_records_to_file(deleted_records, deleted_output_file)
+    transformer._write_output_files(output_file)
```
ghukill, line 44: Some thoughts on `._write_output_files()`.

First of all, I think moving this logic to a method is the right call, so nice work there.

I might also suggest that it should be a **public** method, given it's expected to be called directly from a `Transformer` instance (as opposed to a public method calling private methods not meant for direct usage).

By that same token, I would propose making `write_timdex_records_to_json_file()` and `write_deleted_records_to_txt_file()` **private** methods, as I don't think they should be called directly.

And this brings me to a third point, something I'm pondering in the GeoHarvester as well.  Unceremoniously, but importantly, inside of `write_timdex_records_to_json_file()` there is [this line](https://github.com/MITLibraries/transmogrifier/blob/GDT-83-jsontransformer/transmogrifier/sources/transformer.py#L101) which is actually drivingg the main loop of transformations:
```python
while records:
    ...
```

What I'm not sure is communicated in a method name like `write_output_files()` is that the work of iterating through all input records, transforming them, and _then_ writing is actually getting invoked via this method.

I might suggest that just a small renaming to `transform_and_write_output_files()` would be enough to suggest that not only are files getting written, but this method is kicking off the actual transformation work as well.

transmogrifier/sources/transformer.py
```diff
@@ -72,7 +70,20 @@ def __next__(self) -> TimdexRecord:
                 continue
 
     @final
-    def write_timdex_records_to_json(self, output_file: str) -> int:
+    def _write_output_files(self, output_file: str) -> None:
```
ghukill, line 73: If this method is turned into a public method, I think a docstring here would be nice to also inform the reader that this method is also driving the iteration through all input records, and transformation of them, via the `while records:...` bit.

transmogrifier/sources/transformer.py
```diff
@@ -153,233 +208,308 @@ def parse_source_file(cls, source_file: str) -> Iterator[JSON | Tag]:
...
         Args:
             source_record: A single source record.
         """
-        return []
+        pass
 
     @classmethod
-    @abstractmethod
-    def get_source_record_id(cls, source_record: JSON | Tag) -> str:
+    def get_main_titles(cls, source_record: JSON | Tag) -> list[str]:
```
ghukill, line 235: Should this also be an `@abstractmethod`, requiring it's either defined by `XMLTransformer`, `JSONTransformer`, or their children assuming those two don't define it?

transmogrifier/sources/transformer.py
```diff
@@ -153,233 +208,308 @@ def parse_source_file(cls, source_file: str) -> Iterator[JSON | Tag]:
...
 
         Args:
-            source_record_id: Record identifier for the source record.
             source_record: A single source record.
         """
-        return ""
+        pass
 
-    @classmethod
-    @abstractmethod
```
ghukill, line 245: This feels correct to me, having `get_optional_fields()` **not** be required by any children of this base `Transformer` class, meaning they will receieve an empty dictionary of optional fields if they don't define this method.

transmogrifier/sources/transformer.py
```diff
@@ -153,233 +208,308 @@ def parse_source_file(cls, source_file: str) -> Iterator[JSON | Tag]:
...
         Args:
-            source: Source name.
-            source_record_id: Record identifier for the source record.
-            source_record: A single source record.
+            source_file: A file containing source records to be transformed.
         """
-        return ""
+        with open(source_file, "rb") as file:
+            for record in json.load(file):
+                yield record
```
ghukill, line 334: Don't think this is required now, but noting for the future: I continue to think it might be nice if Transmogrifier could handle [JSON Lines](https://jsonlines.org/) as an input file type.  

This would allow "harvesters" (broadly speaking) to write JSON Lines files for Transmog to work with.

Again, not required, but noting that it could be in this `parse_source_file` where that functionality is introduced later.

ghukill (changes requested):
Overall, I think this looks great.  Thanks for the discussions along the way, as I can see it reflected here.

Curious your thoughts @ehanson8 on the comment around the public vs private nature, and naming, of method `_write_output_files()`.  I believe this code would function as-is, but feels like the right time to nail down the naming of these methods and how we should expect to interact with and understand them.  

While open to pushback on the comments made, requesting changes as I think bare minimum we might want some docstrings on those methods regardless.

But overall, looking good!

transmogrifier/sources/transformer.py
ehanson8: Noted, I'll save that for later!

transmogrifier/cli.py
ehanson8: Agree with all of this, great suggestions!

transmogrifier/sources/transformer.py
ehanson8: Good catch!

ghukill (approved):
New changes look good.

PR #113, opened 2024-01-11
Gdt 116 update locations

ehanson8:
Includes several other minor bug fixes related to recently discovered issues with the `MITAardvark` transform.

Run the following command to run the transform on the fixture:
```
```
Review `output/gismit.json` to see that WKT values are mapped to `locations > geoshape`.

* https://mitlibraries.atlassian.net/browse/GDT-116

#### Developer

#### Code Reviewer

YES


transmogrifier/sources/json/aardvark.py
```diff
@@ -309,14 +304,22 @@ def get_locations(
...
+                        kind=kind_value,
+                    )
+                )
+            else:
                 message = (
-                    f"Geometry field '{aardvark_location_field}' found, but "
-                    f"currently not mapped."
+                    f"Record ID '{source_record_id}': "
+                    f"Unable to parse geodata string '{geodata_string}' "
+                    f"in '{aardvark_location_field}'"
```
ghukill, line 320: Sat on this for awhile, but I think it looks good.  Maybe we'll get into more WKT string types going forward, but for now I think we can be relatively confident they are all `ENVELOPE(...)` strings.  And, therefore, if those fields are present and _don't_ have that `ENVELOPE` string, then it feels accurate we cannot "parse" it.

ghukill (changes requested):
Overall, think it's looking great.  I requested changes mostly for the namespacing by way of semicolons `;` in the subjects.  I did leave a couple questions, but nothing very blocking.  Nice work!  Getting into the fiddly bits of the transformation.

tests/conftest.py
```diff
@@ -163,7 +163,7 @@ def timdex_record_all_fields_and_subfields():
             timdex.Location(
                 value="A point on the globe",
                 kind="Data was gathered here",
-                geodata=[-77.025955, 38.942142],
+                geoshape="BBOX(-77.025955, 38.942142)",
```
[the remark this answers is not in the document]
ehanson8: I'll update it just to be safe, good catch!

tests/sources/json/test_aardvark.py
```diff
@@ -255,11 +259,11 @@ def test_aardvark_get_rights_success(aardvark_record_all_fields):
...
-        timdex.Subject(value=["Country"], kind="DCAT Keyword"),
-        timdex.Subject(value=["Political boundaries"], kind="DCAT Theme"),
-        timdex.Subject(value=["Some city, Some country"], kind="Dublin Core Spatial"),
-        timdex.Subject(value=["Geography"], kind="Dublin Core Subject"),
-        timdex.Subject(value=["Earth"], kind="Dublin Core Subject"),
+        timdex.Subject(value=["Country"], kind="DCAT; Keyword"),
+        timdex.Subject(value=["Political boundaries"], kind="DCAT; Theme"),
+        timdex.Subject(value=["Some city, Some country"], kind="Dublin Core; Spatial"),
+        timdex.Subject(value=["Geography"], kind="Dublin Core; Subject"),
+        timdex.Subject(value=["Earth"], kind="Dublin Core; Subject"),
```
ghukill, line 266: I'm digging it.  I've seen semicolons used for dilineating namespaces in freetext values like this.  Feels like a decent middle ground.

transmogrifier/sources/json/aardvark.py
```diff
@@ -90,9 +90,14 @@ def record_is_deleted(cls, source_record: dict) -> bool:
...
-            return True
-        return False
+        if isinstance(source_record["gbl_suppressed_b"], bool):
+            return source_record["gbl_suppressed_b"]
+        else:
+            message = (
+                f"Record ID '{cls.get_source_record_id(source_record)}': "
+                "'gbl_suppressed_b' value is not a boolean"
+            )
+            raise ValueError(message)
```
ghukill, line 100: I think this is good.  If we don't know definitively if the record is deleted or not, raises an error.

ghukill (approved):
Looks good to me!  Thanks for the updates here.

tests/conftest.py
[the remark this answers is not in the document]
jonavellecuerdo: @ehanson8 I think the space still needs to be added 🤔

tests/test_models.py
```diff
@@ -181,10 +181,10 @@ def test_timdex_record_all_fields_and_subfields(timdex_record_all_fields_and_sub
         timdex_record_all_fields_and_subfields.locations[0].kind
         == "Data was gathered here"
     )
-    assert timdex_record_all_fields_and_subfields.locations[0].geodata == [
-        -77.025955,
-        38.942142,
-    ]
+    assert (
+        timdex_record_all_fields_and_subfields.locations[0].geoshape
+        == "BBOX(-77.025955, 38.942142)"
```
jonavellecuerdo, line 186: Space between `"BBOX"` and opening parentheses `(`?

tests/test_models.py
```diff
@@ -305,7 +305,7 @@ def test_record_asdict_includes_all_fields(timdex_record_all_fields_and_subfield
         "literary_form": "nonfiction",
         "locations": [
             {
-                "geodata": [-77.025955, 38.942142],
+                "geoshape": "BBOX(-77.025955, 38.942142)",
```
jonavellecuerdo, line 308: Space between `"BBOX"` and opening parentheses `(`?

jonavellecuerdo (changes requested):
Hiya' Eric! Small request for change re: syntax for `BBOX (...)` strings in tests to align with [MITAardvark.get_locations() method](https://github.com/MITLibraries/transmogrifier/blob/gdt-116-update-locations/transmogrifier/sources/json/aardvark.py#L316-L318). Might be safer to include that space between characters in the event that OpenSearch requires it. 

Also had one clarifying question!

Otherwise, I think this is looking great. :)

tests/test_models.py
ehanson8: Sorry, I discovered that it would still index without the space, I had shared that with Graham on Slack but I should have noted it on the PR as well!

transmogrifier/helpers.py
```diff
@@ -75,31 +75,6 @@ def parse_date_from_string(
     return None
 
 
-def parse_geodata_string(geodata_string: str, source_record_id: str) -> list[float]:
```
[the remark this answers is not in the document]
ehanson8: Since we're just taking the WKT string as is given than OpenSearch can handle it, we don't need this function to do the parsing that we originally though was needed. Does that make sense?

tests/test_models.py
jonavellecuerdo: Ah, good to know! :) Thank you for clarifying.

jonavellecuerdo (approved):

PR #114, opened 2024-01-24
Update MITAardvark 'rights' field method to support access filter

jonavellecuerdo:

The planned UI mockup for the updated TIMDEX UI includes a feature to filter records
based on "Access to files". This requires a static, aggregatable string value to be present 
in a TIMDEX record that can be used by the TIMDEX API to retrieve relevant records
from OpenSearch.

Review updated `test_aardvark_get_rights_success`: [tests/sources/json/test_aardvark.py](https://github.com/MITLibraries/transmogrifier/blob/GDT-138-update-timdex-record-to-enable-access-filter/tests/sources/json/test_aardvark.py#L248-L299). 

Run `make test` and confirm all tests are passing.

NO

YES

https://mitlibraries.atlassian.net/browse/GDT-138

### Developer

### Code Reviewer(s)


transmogrifier/sources/json/aardvark.py
```diff
@@ -349,17 +349,41 @@ def get_publication_information(source_record: dict) -> list[str]:
...
+                            description="Requires MIT authentication",
+                            kind=kind_access_to_files,
+                        )
+                    )
+                else:
+                    rights.append(
+                        timdex.Rights(
+                            description="Free/open-access", kind=kind_access_to_files
+                        )
+                    )
```
ghukill, line 385: A couple observations:

1. First, I think the logic based on source + `dct_accessRights_s` looks good!  Removed from other comments, this was my understanding of how those values are determined.

2. Given that `dct_accessRights_s` is a **required** field for MITAardvark ([link to JSONSchema in GeoHarvester](https://github.com/MITLibraries/geo-harvester/blob/main/harvester/records/schemas/mit-schema-aardvark.json#L135)), this feels safe to have this nested under line #356 `if "dct_accessRights_s" in source_record` as we know they will always have it.  

My concern -- and it's not specific to this PR -- is how does someone looking at this code alone know that?  Curious @ehanson8's thoughts here, but what if this `get_rights()` method threw an exception if `dct_accessRights_s` is not present?  Then we could dedent everything under line #356 as we'd know it's present.  But, perhaps this is a can of worms as it would suggest similar treatment for other required fields.  And/or, make it brittle in the event we change what's required.  

In summary to #2: probably just fine as-is, but just noting what feels like an unresolved tension with checks for required MITAardvark fields.

3. If `gismit` and `gisogm` were broken out into their own transformations, we also avoid this `if/else` logic.  But again, not in scope for this PR, but just noting as we begin to get into more OGM record considerations.

ghukill (commented):
Looking good!  Just left a comment about the logic (which looks good) and it's relationship to required fields, and `gismit` vs `gisogm` records.  But as stated in the comment, likely out of scope for this PR.

tests/fixtures/aardvark/aardvark_record_all_fields.jsonl
```diff
@@ -1 +1,4 @@
...
...TONWATER95.source.fgdc.xml\"}, {\"label\": \"Normalized Metadata\", \"protocol\": \"Download\", \"url\": \"https://example.com/GISPORTAL_GISOWNER01_BOSTONWATER95.normalized.aardvark.json\"}, {\"label\": \"Data Zipfile\", \"protocol\": \"Download\", \"url\": \"https://example.com/GISPORTAL_GISOWNER01_BOSTONWATER95.zip\"}]}", "dct_rights_sm": ["Some person has the rights"], "dct_rightsHolder_sm": ["The person with the rights", "Another person with the rights"], "dct_spatial_sm": ["Some city, Some country"], "dct_subject_sm": ["Geography", "Earth"], "dct_temporal_sm": ["1943", "1979"], "dct_title_s": "Test title 1", "gbl_dateRange_drsim": ["[1943 TO 1946]"], "gbl_displayNote_sm": ["Danger: This text will be displayed in a red box","Info: This text will be displayed in a blue box","Tip: This text will be displayed in a green box","Warning: This text will be displayed in a yellow box","This is text without a tag and it will be assigned default 'note' style"], "gbl_indexYear_im": [1943,1944,1945,1946], "gbl_resourceClass_sm": ["Dataset"], "gbl_resourceType_sm": ["Vector data"], "gbl_suppressed_b": false, "locn_geometry": "ENVELOPE(-111.1, -104.0, 45.0, 40.9)", "schema_provider_s": "MIT"}
```
ehanson8, line 1: I would prefer we leave this fixture alone and instead use this pattern from some of the other tests to reset the value for the other use cases

```
aardvark_record = next(aardvark_record_all_fields)
aardvark_record["gbl_suppressed_b"] = "True"
```

tests/sources/json/test_aardvark.py
```diff
@@ -256,6 +256,48 @@ def test_aardvark_get_rights_success(aardvark_record_all_fields):
         ),
     ]
 
+    aardvark_record_all_fields_mit_restricted = next(aardvark_record_all_fields)
```
ehanson8, line 259: These should be separate tests, more and smaller tests are much more preferable to me than an aggregate test that has several reasons to fail re: Clean Code

transmogrifier/sources/json/aardvark.py
```diff
@@ -349,17 +349,41 @@ def get_publication_information(source_record: dict) -> list[str]:
...
 
     @staticmethod
-    def get_rights(source_record: dict) -> list[timdex.Rights]:
+    def get_rights(source: str, source_record: dict) -> list[timdex.Rights]:
         """Get values from source record for TIMDEX rights field."""
         rights = []
 
         if "dct_accessRights_s" in source_record:
+            kind_access_rights = "Access"
+            kind_access_to_files = "Access to files"
```
ehanson8, line 358: I get `"Access to files"` but I'm unclear about the difference here, what is `"Access"` referring to? And it looks like `kind_access_rights` is only used once so does it need to be a variable?

transmogrifier/sources/json/aardvark.py
ehanson8: > My concern -- and it's not specific to this PR -- is how does someone looking at this code alone know that? Curious @ehanson8's thoughts here, but what if this `get_rights()` method threw an exception if `dct_accessRights_s` is not present? Then we could dedent everything under line #356 as we'd know it's present. But, perhaps this is a can of worms as it would suggest similar treatment for other required fields. And/or, make it brittle in the event we change what's required.

That feels like overkill if `geo-harvester` is doing any validation around this. We have to do that for outside sources we don't control (RDI sources e.g. WHOAS, Zenodo) but when we know `geo-harvester` considers it required, I don't think it's worth the added complexity of checking again

ehanson8: > 3\. If `gismit` and `gisogm` were broken out into their own transformations, we also avoid this `if/else` logic.  But again, not in scope for this PR, but just noting as we begin to get into more OGM record considerations.

I think that is likely at this point gives the `get_source_link` logic but fine to leave the if/else here for now

transmogrifier/sources/json/aardvark.py
```diff
@@ -349,17 +349,41 @@ def get_publication_information(source_record: dict) -> list[str]:
...
-                    description=source_record["dct_accessRights_s"], kind="Access"
+                    description=source_record["dct_accessRights_s"],
+                    kind=kind_access_rights,
                 )
             )
 
+            if source == "gisogm":
+                rights.append(
+                    timdex.Rights(
+                        description="Not owned by MIT", kind=kind_access_to_files
```
ehanson8, line 369: Curious to get GIS staff feedback on phrasing, I don't know what is expected here (or if it only matters as a facet)

transmogrifier/sources/json/aardvark.py
jonavellecuerdo: Hi @ghukill ! Thank you for your feedback and pointing out **2**. 

I did wonder about this and actually wanted to remove the condition `if "dct_accessRights_s" in source_record` (but forgot to do so). I did get a little confused as to why "rights" wasn't declared a required variable (if it's required in `MITAardvark`), _but_ remembered that TIMDEX doesn't only deal with geospatial metadata.

If we removed the conditional in `line 356`, I think the syntax of `source_record[<var>]` is sufficient in communicating to a programmer that "<var> is required in `source_record` because it is **indexing a value from the dictionary**". 🤔 

In terms of throwing an exception, I'm not sure if this makes sense, but does it make sense to...

I'll try to see what applying the suggestion in **3** looks like and we can see how it feels? 🤓

ehanson8 (commented):
I like the approach, just some questions and suggestions about testing and fixtures

transmogrifier/sources/json/aardvark.py
ehanson8: I think it is likely that we may have `MITAardvark` and `OGMAardvark` classes since we hardcoded a `gismit` string into `MITAardvark.get_source_link()` but we may find a better solution that avoids that

transmogrifier/sources/json/aardvark.py
ghukill: Erg, my comment got deleted when I had to switch contexts.  Will try and keep it brief here.

I agree that explicit checking + exception throwing is overkill, for the reason that GeoHarvester + JSONSchema is ensuring that `dct_accessRights_s` MUST be present.  

To summarize all points, I'd propose:
1. remove `if "dct_accessRights_s" in source_record...` check
2. do not throw exception; handled by direct dictionary access which will throw an exception (but should never happen)

That addresses the majority of my observation(s) above.  What I think this achieves, is that then this newly added "Access to files" logic -- just by looking at code -- is gauranteed to be applied to every record for this `aardvark` transformation, which is what we need and want.

ehanson8: 💯 Totally agree with Graham's proposal

tests/sources/json/test_aardvark.py
jonavellecuerdo: Change has been made. Please see the latest commit.

tests/fixtures/aardvark/aardvark_record_all_fields.jsonl
jonavellecuerdo: Change has been made. See the latest commit. Thank you for this suggestion! Definitely cleaner. 😎

transmogrifier/sources/json/aardvark.py
jonavellecuerdo: I am not sure of the intention of `kind = "Access"`, but all it seems to do is retrieve the value from the Aardvark field [dct_accessRights_s](https://opengeometadata.org/ogm-aardvark/#access-rights). For the MITAardvark transformer, the `get_rights()` field method essentially pulls from fields under the [Aardvark Rights category](https://opengeometadata.org/ogm-aardvark/#rights).

@ehanson8 and @ghukill - What do you think of updating `kind="Access"` -> `kind="Access rights"`?

ehanson8:
Changes look good but still curious about [this](https://github.com/MITLibraries/transmogrifier/pull/114#discussion_r1465384128)

transmogrifier/sources/json/aardvark.py
ehanson8: Maybe a better way of stating my question is: is there actually a difference between `Access` and `Access to files`? There may not be but if we have 2 different strings, the user may think there's a difference

ghukill: I agree with @ehanson8 about "Access" vs "Access to files" being very similar and potentially confusing.  But the language for the values, and the language for that kind both come directly from the UI mockup:

7328473a-bcf5-4d91-9fbb-6458922eef72

While maybe we could make the kind something like "Download Access", we'd need to confirm with EngX that the actual "kind" value could differ from the text on the UI.  I imagine it could, but unsure.

OR, we could just continue with "Access to files" (which does _kind_ of suggest downloading), and then iterate later if determined confusing.

As far as the string values go (e.g. "Not owned by MIT"), I'd imagine those are good as-is because they _will_ render directly on the page via the aggregation.

jonavellecuerdo:
@ehanson8 @ghukill Let me know what you think of the latest changes!

1. The UI mockup indicates "MIT Authentication", but I propose "Required MIT authentication" so that all string values will be adjectives. 🤓 
2. Acknowledging that the `if-else` conditions in `MITAardvark.get_rights()` field method may not be needed once we create an `OGMAardvark` transformer.
3. We will circle back on `Rights.kind` values and will check in with EngX (@JPrevost) on whether or not the values in the TIMDEX record can (or should) differ from the TIMDEX (Geoweb) UI.

JPrevost:
@jonavellecuerdo I haven't read through the full discussion here, but is there a transform in dev1 S3 that has been created with this branch so I can pull in the data to better understand the ramifications?

transmogrifier/sources/json/aardvark.py
ehanson8: I'm OK with that as is given the draft status and I get it now seeing the UI, but to me `Access to files` seems like the only type of access users would care about. But this is better sorted out in inter-team discussions later!

JPrevost: Some context (this may no longer be relevant, I haven't checked to see what you all decided on in the code).

In general, field values _can_ be manipulated between OpenSearch and TIMDEX API (or TIMDEX API and TIMDEX UI) when we have a good reason to do so, but when we find ourselves wanting to do that we should probably pause and ask why. If UX is suggesting a label in the UI is the best way to describe something from a user perspective, is there a reason an API user would not find the same language helpful? Or is there a reason we want to introduce potential future confusion between what is in OpenSearch and TIMDEX API?

So before forking our language across tiers we should reach out to UX and explain the situation and ask if they would consider language we feel is more specific or accurate. If there is a good reason to make a distinction between how we refer to data in TIMDEX Spec and TIMDEX API/TIMDEX UI, we can definitely support it... we just need to be sure that's what we want :)

ghukill: I better see your point @ehanson8, that _conceptually_ what's even the difference...  And, agreed about inter-team discussing, which [this slack thread might tease out](https://mit-libraries.slack.com/archives/C05UK9PDNBW/p1706195251372659).

AND, one of the open questions for GDT-138 is whether this will even be a `Rights` entry, or maybe a dedicated field, so might all be moot if it's a new field.

jonavellecuerdo:
@JPrevost Thank you for asking! Here is an example of a transform in dev1 S3 with the proposed changes: [JSON file](https://s3.console.aws.amazon.com/s3/object/timdex-extract-dev-222053980223?region=us-east-1&bucketType=general&prefix=gismit/gismit-2024-01-03-full-transformed-records-to-index-ACCESS-FILTER-TEST.json). Also, I'll be sending over some notes from the spike work investigating **nested, filtered aggregations in OpenSearch** (heads up -- yes we can 😄 ).

jonavellecuerdo:
@ehanson8 @ghukill 

Please see the latest commit. Here are the main takeaways:

1. Ultimately, we confirmed that the proposed solution of introducing a new instance of "Rights" to `TimdexRecord.rights` field to support "Access to file" filters is [(a) technically achievable with openSearch](https://mitlibraries.atlassian.net/browse/GDT-138?focusedCommentId=122738) and [(b) it is reasonable to use this field in this manner, as exemplified by at least one other transformer](https://mitlibraries.atlassian.net/browse/GDT-138?focusedCommentId=122738).

2. During one of our sync-up meetings, @ghukill and I thought that a solution to address the comment on "Access" vs. "Access to files" could be to update the code such that:

```
timdex.Rights(description=<geoharvester.harvester.records.MITAardvark.dct_accessRights_s>, kind="Access rights")
``` 
3. I decided to update "Requires MIT authentication" -> "MIT authentication" to align the value with the proposed UI's display. 

Let me know what you think!

transmogrifier/sources/json/aardvark.py
```diff
@@ -385,16 +385,36 @@ def get_publication_information(source_record: dict) -> list[str]:
         return publication_information
 
     @staticmethod
-    def get_rights(source_record: dict) -> list[timdex.Rights]:
+    def get_rights(source: str, source_record: dict) -> list[timdex.Rights]:
         """Get values from source record for TIMDEX rights field."""
         rights = []
+        kind_access_to_files = "Access to files"
```
ehanson8, line 391: The variable name is longer than the actual string so it feels unnecessary to use a variable

ehanson8 (approved):
One suggested change

transmogrifier/sources/json/aardvark.py
ehanson8: On second thought, that's fine. Disregard my comment

PR #116, opened 2024-01-30
Linting update

ehanson8:
The following documentation related error codes were ignored as they were numerous but did not affect application functionality: 

- D212 https://docs.astral.sh/ruff/rules/multi-line-summary-first-line/ 
- D402 https://docs.astral.sh/ruff/rules/no-signature/ 
- D205 https://docs.astral.sh/ruff/rules/blank-line-after-summary/ 

Responded to most linter suggestions and placed 1 line ignores where the changes would compromise readability.

Declined to install type stubs for `BeautifulSoup` as it introduced 168 `mypy` errors and this application has been running fine in prod for some time now. 

YES

NO

### Developer

### Code Reviewer(s)
- [x] The commit message is clear and follows our guidelines (not just this PR message)
- [x] The provided documentation is sufficient for understanding any new functionality introduced
- [x] Any manual tests have been performed and verified


transmogrifier/config.py
```diff
@@ -1,11 +1,12 @@
...
+
 import json
 import logging
 import os
-from typing import Literal, Union
+from typing import Literal
 
 import sentry_sdk
-from bs4 import BeautifulSoup
+from bs4 import BeautifulSoup  # type: ignore[import-untyped]
```
ghukill, line 9: I think this was a good call to not install this type library.

transmogrifier/config.py
```diff
@@ -173,7 +174,8 @@ def load_external_config(
     with open(file_path, "rb") as config_file:
         if file_type == "json":
             return json.load(config_file)
-        elif file_type == "xml":
+        elif file_type == "xml":  # noqa: RET505
             return BeautifulSoup(config_file, "xml")
         else:
-            raise ValueError("Unrecognized file_type parameter: %s", file_type)
+            message = f"Unrecognized file_type parameter: {file_type}"
+            raise ValueError(message)
```
ghukill, line 181: I think this was a good call too.  I looked at the code as-is, and I agree this is the most readable form.  Sure, we might get away with removing an `elif` or `else` here and there, but as-is it reads very clearly.

transmogrifier/helpers.py
```diff
@@ -69,7 +68,7 @@ def parse_date_from_string(
     """
     for date_format in DATE_FORMATS:
         try:
-            return datetime.strptime(date_string, date_format)
+            return datetime.strptime(date_string, date_format).astimezone(UTC)
```
ghukill, line 71: I don't think we need to worry about the repercussions of this, but noting this could affect the timestamps derived.  If we had logic based on these timestamps, or if they were financial/audit/etc. information I think it'd be a bigger deal.  But given this is just parsing date strings from metadata, a difference of a couple of hours from previous runs seems unimportant.

transmogrifier/helpers.py
```diff
@@ -120,31 +118,29 @@ def validate_date_range(
...
+        "Record ID '%s' has invalid values in a date range: '%s', '%s'",
+        source_record_id,
+        start_date,
+        end_date,
+    )
+    return False
 
 
-class DeletedRecord(Exception):
+class DeletedRecordError(Exception):
```
ghukill, line 137: I've been making this switch of trying to add "Error" as well since moving to Ruff.  Most of the time, I generally like it, but I wonder if it makes sense in this situation.

Reading a bit about python exceptions, I think there is case to be made that sometimes exceptions are used for **flow control**, which I feel is the case here.

I understand the linting suggestion to help make it clear that this class is an exception, and not get it confused for something else, which makes sense.  What about something like `DeletedRecordEvent`?  To me, the "Event" suffix still aligns with raising it, plus handling it,  but it distinguishes it from some kind of error.  My feeling is that we correctly understand that a record is deleted, but raising an exception is just a tidy way of sending an **event** that should be handled higher in the application logic and flow.

transmogrifier/models.py
```diff
@@ -1,25 +1,31 @@
...
-        raise ValueError(
+        message = (
             f"{attribute.name} may have a 'lt' or 'lte' value, but not both; "
             f"received {value}"
         )
+        raise ValueError(message)
 
 
-def list_of(item_type):
+def list_of(item_type: Any) -> Callable:  # noqa: ANN401
```
ghukill, line 28: I think `Any` is a good call here. Sure, we could scan the codebase and create a fairly large union of possible types... but we don't gain much from that.

transmogrifier/sources/json/aardvark.py
```diff
@@ -225,17 +227,14 @@ def _range_dates(
...
-        ]:
+        for date_range_string in source_record.get("gbl_dateRange_drsim", []):
             date_range_values = cls.parse_solr_date_range_string(
                 date_range_string, source_record_id
             )
             range_dates.append(
                 timdex.Date(
                     kind="Coverage",
-                    range=timdex.Date_Range(
+                    range=timdex.DateRange(
```
ghukill, line 237: I imagine this change may have been a pain to track down, but I think it's a good one!

transmogrifier/sources/json/aardvark.py
```diff
@@ -417,6 +417,8 @@ def get_subjects(source_record: dict) -> list[timdex.Subject]:
             if key in source_record
         }.items():
             for subject in source_record[aardvark_subject_field]:
-                subjects.append(timdex.Subject(value=[subject], kind=kind_value))
+                subjects.append(  # noqa: PERF401
+                    timdex.Subject(value=[subject], kind=kind_value)
+                )
```
ghukill, line 422: I was digging into the `PERF401` linting ignore, and found myself still tripping a bit over this method.  

I think it's a sneaky one, as it's doing quite a bit: looping through possible field names, that may or may not have data, and setting an associated kind value if present.

Wanted to float an alternate approach that makes use of `.extend()` and `.get()` to avoid some checks, and allows for list comprehension.  Not a requested change, just a suggestion for an alternate form that I don't think requres linting ignores.  Hard to say if it's more or less readable.

```python
subjects = []

aardvark_subject_fields = {
    "dcat_keyword_sm": "DCAT; Keyword",
    "dcat_theme_sm": "DCAT; Theme",
    "dct_spatial_sm": "Dublin Core; Spatial",
    "dct_subject_sm": "Dublin Core; Subject",
    "gbl_resourceClass_sm": "Subject scheme not provided",
    "gbl_resourceType_sm": "Subject scheme not provided",
}

for subject_field, subject_kind in aardvark_subject_fields.items():
    subjects.extend(
        [
            timdex.Subject(value=[subject], kind=subject_kind)
            for subject in source_record.get(subject_field, [])
        ]
    )

return subjects
```

transmogrifier/sources/transformer.py
```diff
@@ -1,27 +1,31 @@
...
 # should not be a security issue.
 from lxml import etree  # nosec B410
 
 from transmogrifier.config import SOURCES
-from transmogrifier.helpers import DeletedRecord, generate_citation
+from transmogrifier.helpers import DeletedRecordError, generate_citation
 from transmogrifier.models import TimdexRecord
 
+if TYPE_CHECKING:
+    from collections.abc import Iterator
```
ghukill, line 27: Nice!  I'm warming up to the `Iterator` type quite a bit.

transmogrifier/sources/transformer.py
```diff
@@ -111,18 +111,21 @@ def _write_timdex_records_to_json_file(self, output_file: str) -> int:
             while record:
                 file.write(
                     json.dumps(
-                        asdict(record, filter=lambda attr, value: value is not None),
+                        asdict(
+                            record,
+                            filter=lambda attr, value: value is not None,  # noqa: ARG005
+                        ),
```
ghukill, line 117: Just a suggestion: could use a leading `_` underscore to indicate the `attr` variable is not needed, and then remove the linting ignore:

```python
asdict(record, filter=lambda _attr, value: value is not None),
```

Or arguably even more explicit:
```python
asdict(record, filter=lambda _, value: value is not None),
```

transmogrifier/sources/transformer.py
```diff
@@ -243,26 +242,26 @@ def _transform(
...
+        fields = {
+            **self.get_required_fields(source_record),
+            **optional_fields,
+        }
+
+        # If citation field was not present, generate citation from other fields
+        if fields.get("citation") is None:
+            fields["citation"] = generate_citation(fields)
+        if fields.get("content_type") is None:
+            fields["content_type"] = ["Not specified"]
```
ghukill, line 259: This is a case where I think the removed `else` improves readability; kind of sets the default route as the happy path.

transmogrifier/sources/transformer.py
```diff
@@ -627,7 +625,7 @@ def get_source_link(
 
     @classmethod
     def get_timdex_record_id(
-        cls, source: str, source_record_id: str, source_record: Tag
+        cls, source: str, source_record_id: str, source_record: Tag  # noqa: ARG003
```
ghukill, line 628: I landed on this one for a comment, but I think it's applicable to any of the methods that have an argument that is defined on the parent class, but not used in the child classes.  However, only works when the argument is **positional**; don't think it'd be as simple for named arguments.

We could also lean on the leading underscore to indicate the argument is not used and thereby avoid linting ignores:

```python
cls, source: str, source_record_id: str, _source_record: Tag
```

This is not a requested change, just noting.  This would likely require multiple changes, and I'm not sure if it actually improves readability.  Although, if one found themselves updating the codebase and wanted to use the variable later, they might pause with the leading underscore and think to themselves it wasn't used previously.  

All in all, probably not worth the change, but wanted to note.

ghukill (changes requested):
All in all, heroic work here!  Thanks for taking this on.  The vast, vast majority to me looks outstanding.  I can already tell that hopping into this codebase will be a bit more inviting now, given the increased typing and linting help.

I left a few optional suggestions, and would have marked this PR as approved with those as just suggestions, but the one I think warrants some discussion is the exception `DeletedRecord` vs `DeletedRecordError` ([comment link](https://github.com/MITLibraries/transmogrifier/pull/116#discussion_r1472901825)). 

That said, great work.  Looking forward to getting this merged!

transmogrifier/helpers.py
ehanson8: Good point and agree it's probably not as big of a deal here as opposed to `alma-creditcardslips` or `alma-sapinvoices`

transmogrifier/helpers.py
ehanson8: I felt a little weird about adding `Error` this one but you summed up the reason why, I think `DeletedRecordEvent` is much more accurate

transmogrifier/sources/json/aardvark.py
ehanson8: This is much more readable, thanks!

transmogrifier/helpers.py
```diff
@@ -134,7 +134,7 @@ def validate_date_range(
     return False
 
 
-class DeletedRecordError(Exception):
+class DeletedRecordEvent(Exception):  # noqa: N818
```
ghukill, line 137: 👍

ghukill (approved):
Looks good!  Thanks for the considerations and quick updates.

transmogrifier/sources/transformer.py
ehanson8: A little bit of work but necessary, fixed!

PR #117, opened 2024-02-06
Add description to readme

ehanson8:
Adding a high-level overview of the app to `README.md`. This went through several drafts, my intended audience was a new engineer joining the team who needed a basic understanding of the app before diving into the code. Let me know what should be added/deleted/rephrased. 

YES

NO

### Developer

### Code Reviewer(s)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,36 @@
 # transmogrifier
 
-Data transformation CLI app for TIMDEX
+An application to transform source records to the TIMDEX data model to facilitate ingest into an OpenSearch index.
+
+## Description
+
+TIMDEX ingests records from various sources with different metadata formats, necessitating an application to transform those source records to a common metadata format. This application processes both XML and JSON source records and outputs a JSON file of records formatted according to the TIMDEX data model. 
+
+```mermaid
+---
+title: Transmogrifier in the TIMDEX pipeline
+---
+flowchart TD
+    ArchivesSpace
+    DSpace
+    GeoData
+    MARC
+    transmogrifier((transmogrifier))
+    JSON
+    timdex-index-manager
+    ArchivesSpace[("ArchivesSpace\n(EAD XML)")] --> transmogrifier
+    DSpace[("DSpace\n(METS XML)")] --> transmogrifier
+    GeoData[("GeoData\n(Aardvark JSON)")] --> transmogrifier
+    MARC[("Alma\n(MARCXML)")] --> transmogrifier
+    transmogrifier --> JSON["TIMDEX JSON"]
+    JSON[TIMDEX JSON file] --> timdex-index-manager((timdex-index-manager))
+```
+
+The TIMDEX data model is designed to produce records that can be successfully ingested into an OpenSearch index and contains data fields that are broadly applicable to various types of records. `transmogrifier` contains different validators to ensure that the record is structured properly and that certain types of values, such as dates, align with OpenSearch's expectations.
+
+Each source is defined with configuration values and a dedicated transform class to process records from that source. For each transform class, various errors and warnings are logged. Some errors are logged and the entire source record is skipped because it is incorrectly formatted, while others are merely logged as warnings for later review. The application also determines which records are marked as deleted in each source and removes those record from the OpenSearch index. 
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,6 +1,36 @@
 # transmogrifier
 
-Data transformation CLI app for TIMDEX
+An application to transform source records to the TIMDEX data model to facilitate ingest into an OpenSearch index.
+
+## Description
+
+TIMDEX ingests records from various sources with different metadata formats, necessitating an application to transform those source records to a common metadata format. This application processes both XML and JSON source records and outputs a JSON file of records formatted according to the TIMDEX data model.
```
ghukill, line 7: I don't know how to word it, but I think it could be helpful to draw a straight line between "common metadata format" and "TIMDEX data model", that they are largely one and the same.

Something you're probably wise not to introduce here is the "contract" between Transmog (python class that codifies the data model) and TIM (Opensearch mapping JSON that codifies that data model).  I would argue the data model is the combination and relationship of those two things, but unsure if that complexity belongs here.

README.md
```diff
@@ -1,6 +1,36 @@
...
+    DSpace[("DSpace\n(METS XML)")] --> transmogrifier
+    GeoData[("GeoData\n(Aardvark JSON)")] --> transmogrifier
+    MARC[("Alma\n(MARCXML)")] --> transmogrifier
+    transmogrifier --> JSON["TIMDEX JSON"]
+    JSON[TIMDEX JSON file] --> timdex-index-manager((timdex-index-manager))
+```
+
+The TIMDEX data model is designed to produce records that can be successfully ingested into an OpenSearch index and contains data fields that are broadly applicable to various types of records. `transmogrifier` contains different validators to ensure that the record is structured properly and that certain types of values, such as dates, align with OpenSearch's expectations.
+
+Each source is defined with configuration values and a dedicated transform class to process records from that source. For each transform class, various errors and warnings are logged. Some errors are logged and the entire source record is skipped because it is incorrectly formatted, while others are merely logged as warnings for later review. The application also determines which records are marked as deleted in each source and removes those record from the OpenSearch index.
```
ghukill, line 31: This is some nice writing.  It's covering a _lot_ of ground, but it's clear and concise and easy to follow.  Nice work!  Maybe additional sections or docs are added later, or links out, but this is a great overview of this codebase right in front of you.

ghukill: One suggestion: I wonder if it'd be helpful to say something like "it is skipped because it is does not meet minimum TIMDEX record requirements" instead of "it is incorrectly formatted"?

Thinking about recent discussions leveraging the distinction between format and data.  If a record is fully missing a field, not sure I'd consider that formatting, it feels like it's missing required data.  Maybe just semantics but feels like an opportunity to seed the reader with a question like, "well, where _are_ the TIMDEX data model requirements?"  and not just assume it's a bad XML angle bracket or something that causes records to get skipped.

ghukill (approved):
I think it's a stellar upgrade to the README.  Left a couple of comments, but approved with or without changes.

Think you hit a really nice balance between succintness and clarity.

README.md
ehanson8: I think it's actually implying that something is seriously wrong with the source record, such as EAD without an `archdesc` element or MARC without an `008` or leader field. I can definitely phrase it better but it's closer to bad angle bracket in the source than not having a required field, if that makes sense

README.md
ehanson8: Good call, will do!

README.md
ehanson8: Maybe `Some errors are logged and the entire source record is skipped because the severity implies it should not be processed until fixed, while others are merely logged as warnings for later review.`?

ghukill: > Maybe `Some errors are logged and the entire source record is skipped because the severity implies it should not be processed until fixed, while others are merely logged as warnings for later review.`?

I think that's stellar.  Thanks for the update.  I think the "severity" word is key, as it bypasses tricky conversations of "format" vs "data", technically invalid XML vs XML that doesn't have arbitrarily required fields by us.  Looks good to me!

PR #124, opened 2024-02-14
Validate date strings when transforming MITAardvark records

ghukill:

The primary change this PR introduces is validating date strings for MITAardvark transformations, particularly for records from OpenGeoMetadata (OGM) where dates are less controlled than when we create them.

Using `validate_date(...)` is inline with most other transformations in Transmogrifier.  By adding this, all date related errors while indexing to Opensearch locally were avoided, allowing an additional 26k records to index.  However, this means we also lost adding dates to TIMDEX records that were not immediately valid, e.g.:

- `"1993?"`
- `"approximately 1569"`
- `"[1910 to 1919]"`

While not ideal to lose information during transformation, this seems acceptable at this time for a couple of reasons:

- this approach keeps this transformation inline with other transformations that also employ this validation
- while some dates are lost, it's very rarely not the _only_ date for an OGM record
- when we return to dates in Transmog -- possibly cleaning common patterns like `?` or `xxx to yyy` -- we can apply that logic once and know that all transformations will benefit

**Also in this PR**:
- source `gisogm` should also use the `MITAardvark` transformation as those records; MIT and OGM share transformation
- small tweak to skip multiple download links if present in field `dct_references_s` from Aardvark record

The new unit test in [this commit](https://github.com/MITLibraries/transmogrifier/pull/124/commits/5d166e88b8996358f54cb4919b118a55b428e84d) is indicative of how it drops values that are not valid.

NO

YES: increase in documents indexed, decrease in multiple dates per record

- https://mitlibraries.atlassian.net/browse/GDT-109

### Developer

### Code Reviewer(s)


transmogrifier/config.py
```diff
@@ -112,7 +112,7 @@
     "gisogm": {
         "name": "OpenGeoMetadata GIS Resources",
         "base-url": "https://search.libraries.mit.edu/record/",
-        "transform-class": "transmogrifier.sources.json.aardvark.OGMAardvark",
+        "transform-class": "transmogrifier.sources.json.aardvark.MITAardvark",
```
ehanson8, line 115: So, `OGMAardvark` was referenced here before being defined which we were expecting to need because of the `get_source_link` method including `gismit`, are we thinking differently about that approach now?

ehanson8 (commented):
Looks good but 1 bigger picture question before approving

transmogrifier/config.py
ghukill: Really glad you raised this up, for a couple of reasons.

First of all, looking more closely at the `get_source_link` which I had not thought to check, I believe that we can update that for MITAardvark transformations to use data from `dct_references_s`.  In this way, it also helps explain why MIT and OGM records can share the same transformer.

Every record that comes out of GeoHarvester is an "MIT" Aardvark record in the sense that, regardless of origin institution or metadata format, we have crafted the Aardvark file in a way that meets our TIMDEX needs.  During that work in GeoHarvester, quite a bit of care is taken to craft the `dct_references_s` field which contains URLs.

The value for `dct_refereces_s['http://schema.org/url']` _is_ what the "source link" for the record should be.  For MIT records this will be `https://geodata.libraries.mit.edu/record/<IDENTIFIER>` and for OGM records it will be an external URL that we extracted from the source metadata; gauranteed to be present or it does not get included in the harvester output.

Taking all this together, will work on another commit that:
1. updates `get_source_link` to actually read data from the record
2. will remove `base_url` from the `gismit` and `gisogm` configurations, as it's not needed

ghukill: @ehanson8 - just pushed [this commit](https://github.com/MITLibraries/transmogrifier/pull/124/commits/aa84d194cb1727ddf39844f439fed205308a163f) and have re-requested review.

transmogrifier/sources/json/aardvark.py
```diff
@@ -32,25 +32,33 @@ def get_main_titles(cls, source_record: dict) -> list[str]:
...
         """
-        return source_base_url + cls.get_timdex_record_id(
-            "gismit", source_record_id, source_record
-        )
+        links = cls.get_links(source_record, source_record_id)
+        url_links = [link for link in links if link.kind == "Website"]
+        if len(url_links) == 1:
+            return url_links[0].url
+        message = "Could not locate a kind=Website link to pull the source link from."
+        raise ValueError(message)
```
ehanson8, line 61: Perfect, much better than what we were doing before

ehanson8 (approved):

PR #125, opened 2024-02-15
ADR to handle dataset information in TIMDEX

ghukill:

This PR introduces a new Architecture Decision Record (ADR) that proposes to add a new `form` field to the TIMDEX data model.  

This discussion was kicked off when seeking to understand what TIMDEX fields to use for facet filters in a new geospatial TIMDEX UI.  Reaching a decision on where to store this data in a TIMDEX record will unblock work for DataEng (work to store the data) and EngX (utilize the data for facets).

Rendered markdown for current version of the ADR: https://github.com/MITLibraries/transmogrifier/blob/GDT-193-adr-handle-data-forms/docs/adrs/0002-field-for-data-type-form-information.md

NO

NO

- https://mitlibraries.atlassian.net/browse/GDT-193

### Developer

### Code Reviewer(s)

### DIFF
--- a/docs/adrs/0002-field-for-data-type-form-information.md
+++ b/docs/adrs/0002-field-for-data-type-form-information.md
@@ -231,20 +231,15 @@ might be helpful to decouple that from this decision at hand.
     - e.g. `application/pdf` suggests "PDF", or `text/csv` suggests "CSV", to name a couple example
     - there are python libraries that can handle 90% of these conversions, if a friendly form is not present in the library
 
-Examples:
-```json
-{
-    "content_type": ["Polygon data"],
-    "format": "Shapefile"
-}
-```
+Examples across multiple sources:
 
-```json
-{
-    "content_type": ["Raster data", "Image"],
-    "format": "GeoTIFF"
-}
-```
+| `source`       | `content_class :str` [_proposed but not exists_] | `content_type :list[str]`   | `format :str` | `file_formats :list[str]`      |
+|----------------|--------------------------------------------------|-----------------------------|---------------|--------------------------------|
+| gismit/gisogm  | Geospatial data                                  | [Raster data, Image data]   | GeoTIFF       | [image/tiff]                   |
+| gismit/gisogm  | Geospatial data                                  | [Polygon data, Vector data] | Shapefile     | [application/zipped-shapefile] |
+| alma           | Language material                                | [Short stories]             | EPub Ebook    | [application/epub+zip]         |
+| alma           | Cartographic material                            | [Quadrangle map]            | Scanned map   | [image/jpeg]                   |
+| dspace         | Language material                                | [Academic thesis]           | PDF           | [application/pdf]              |
 
 ## Decision
 
--- a/docs/adrs/0002-field-for-data-type-form-information.md
+++ b/docs/adrs/0002-field-for-data-type-form-information.md
@@ -1,0 +1,162 @@
+# 1. Field for Data Type / Form Information
+
+Follows: https://mitlibraries.github.io/guides/misc/adr.html
+
+Date: 2024-02-15
+
+## Status
+
+Proposed
+
+## Context
+
+### Need
+
+Identify a pre-existing, or new, TIMDEX field to store information about GIS datasets that will support required facet filters, while also potentially serving other TIMDEX data sources.
+
+### Overview
+During a project that introduced geospatial data into TIMDEX, two new sources were added `gisogm` and `gismit`, both of which use the new Transmogrifier transformer 
+`MITAardvark`. Records for these sources come from the [GeoHarvester](https://github.com/MITLibraries/geo-harvester) which produces [Aardvark metadata schema](https://opengeometadata.org/ogm-aardvark/) JSON records.    
+
+During work on a new geospatial TIMDEX UI, the question was raised as to what fields should be used for two new required facet filters to mirror functionality in the [current Geoweb application](https://geodata.mit.edu/):
+
+  * "Data Type"
+    * values like: `[Polygon, Point, Raster, Image, etc.]`
+  * "Format"
+    * values like: `[Shapefile, TIFF, GeoTIFF, JPEG, etc.]`
+
+This revealed a potential gap in the TIMDEX data model.  While the TIMDEX field `format` is a great fit for the "Format" values we expect to see in that facet filter, there does not appear to be 
+an appropriate field to hold values that would populate the "Data Type" facet filter.
+
+It is worth noting that the facet filter "Content Types", driven by the TIMDEX field `content_type`, is set statically for these records as `Geospatial data`.  This allows them to sit next to 
+other TIMDEX records with `content_type` values like `[Language material, Musical sound recording, Thesis, etc.]`, thereby also removing `content_type` as a possible field for the issue at hand. 
+
+Analysis of the source `MITAardvark` records suggests that the Aardvark field `gbl_resourceType_sm` contains the values we'll want for the facet filter "Data Type".  The 
+Aardvark schema defines this field as:
+
+> "...categories for classifying the spatial type or structure of a dataset"
+
+In this way, the TIMDEX data model is lacking a field that can be used to describe the "spatial type or structure of a dataset".  Maybe stated more generally, a field that describes
+the structure of the data _inside_ the format noted in `format`.
+
+It is also worth noting that data values for both facets -- present in fields `dct_format_s` and `gbl_resourceType_sm` from the Aardvark records -- will be controlled values
+when written by the GeoHarvester:
+
+  * "Format" values:
+    * https://opengeometadata.org/ogm-aardvark/#format-values (e.g. "Shapefile", "GeoTIFF", "TIFF" etc.)
+  * "Data Type" values (from either list):
+    * https://opengeometadata.org/ogm-aardvark/#resource-type-values-loc (e.g. "Atlases", "Cadastral maps", "Quadrangle maps")
+    * https://opengeometadata.org/ogm-aardvark/#resource-type-values-ogm (e.g. "Polygon", "Raster", "Line", "Point")
+
+### Possible Solutions
+
+#### Option 1- Use `subjects` with `kind="Data Type"`
+
+In this approach, "Data Type" values would be stored as `subjects` with `kind="Data Type"`.
+
+Example:
+```json
+{
+    "subjects": [
+        {
+            "value": "Polygon",
+            "kind": "Data Type"
+        },
+        {
+            "value": "Vector",
+            "kind": "Data Type"
+        }
+    ]
+}
+```
+
+Pros:
+  * does not require a change to TIMDEX data model anywhere
+
+Cons:
+  * these "Data Type" values don't feel like subjects; they are not really _about_ the resource so much as describing its type/structure/form 
+
+
+#### Option 2- Create new, multivalued string field `form`
+
+In this approach, "Data Type" value would be stored in a new, multivalued string field `form`:
+
+Example:
+```json
+{
+    "form": ["Polygon", "Vector"]
+}
+```
+
+Pros:
+  * purely additive change to data model
+  * simple, top level property makes aggregations very simple
+
+Cons:
+  * still require, and sit along next to, `literary_form` field for describing text sources as "Fiction" or "Nonfiction"
+
+
+#### Option 3- Create new, multivalued objects field `form`; collapse `literary_form` into this
+
+In this approach, "Data Type" value would be stored in a new, multivalued object field `form`:
+
+Example:
+```json
+{
+    "form": [
+        {
+            "value": "Polygon",
+            "kind": "Data Type"
+        },
+        {
+            "value": "Vector",
+            "kind": "Data Type"
+        }
+    ]
+}
+```
+
+Pros:
+  * allows collapsing of `literary_form` field; noting some shared sentiment that this field might be too source-specific for TIMDEX
+  * like other object fields, leaves the door open for adding a `uri` property at a later time
+
+Cons:
+  * would require reworking the transformations + re-indexing any sources that use `literary_form`
+  * nested field type, a bit harder to query for aggregations
+
+## Decision
+
+It is proposed to move forward with **Option 3: Create new, multivalued objects field `form`; collapse `literary_form` into this**. 
+
+This approach would:
+  * readily support the "Data Type" values needed by the new geospatial TIMDEX UI's facet filters
+  * cover values currently stored in the field `literary_form` which may have been too source-specific to begin with
+  * allow future application of URIs
+  * support other "form"-like values we may not have encountered yet, with the flexibility and additional granularity of the `kind` qualifier 
+
+The decision to add a new `form` field would rely on a shared and documented understanding of how `form` differs from `format`.  Proposing a couple
+shorthand distinctions:
+
+  * `format`:
+    * describes the _external_ nature of the resource
+    * e.g. TIFF, Shapefile, PDF, Print book, Electronic resource, Online archival collection, etc.
+  * `form`
+    * describes the _internal_ structure or conceptual arrangement of the resource, regardless of its external format
+    * e.g. Polygon, Image, Raster, Nonfiction, Novel, Textbook, Poetry, Biography, Reference Book, Personal papers, Organizational records, etc.
+
+## Consequences
+
+Adding this new field to the TIMDEX data model would require the following work and changes:
+  * Transmogrifier
+    * `TimdexRecord` class definition
+    * any pre-existing applications of the `literay_form` field
+    * check if `subjects@kind=Genre/Form` applied anywhere, if so, consider `form`
+  * timdex-index-manager (TIM)
+    * Opensearch mapping
+  * re-run an sources that used `literary_form`
+  * dynamically apply mapping update to ALL indexes to include new field `form`
+    * NOTE: this action was needed already for the new field `locations.geoshape`
+  * GraphQL API
+    * add `form` to schema
+    * add to documentation
+    * deprecate `literary_form` field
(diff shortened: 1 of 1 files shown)


docs/adrs/0002-field-for-data-type-form-information.md
```diff
@@ -0,0 +1,162 @@
...
+
+Analysis of the source `MITAardvark` records suggests that the Aardvark field `gbl_resourceType_sm` contains the values we'll want for the facet filter "Data Type".  The 
+Aardvark schema defines this field as:
+
+> "...categories for classifying the spatial type or structure of a dataset"
+
+In this way, the TIMDEX data model is lacking a field that can be used to describe the "spatial type or structure of a dataset".  Maybe stated more generally, a field that describes
+the structure of the data _inside_ the format noted in `format`.
+
+It is also worth noting that data values for both facets -- present in fields `dct_format_s` and `gbl_resourceType_sm` from the Aardvark records -- will be controlled values
```
JPrevost, line 42: I wonder if `ResourceType` is more or less confusing than `Form` for the new object name? I _do_ think `Form` works and I will support it, but the fact that at least one other use case of this type of data refers to it as `ResourceType` seems worth explicitly considering.

JPrevost (commented):
I'm in agreement with the core proposal here. I want to consider if `Form` is the best label for this new Object.

I may have some very small additional details to add some more context and once I have had a chance to do that and we have considered `Form` vs `ResourceType` vs `?` I will feel confident in approving this as-written.

Solid write up, thanks!

docs/adrs/0002-field-for-data-type-form-information.md
ghukill: @JPrevost - do you think `ResourceType` could still take values formerly used in `literary_form`? e.g.

```json
"resource_type":[
  {
    "value":"Nonfiction",
    "kind":"Literary Form"
  }
]
```

I'd also wonder how this would to `content_type` which we [define in the spec](https://mitlibraries.github.io/timdex/reference/#definition-Record) as:

> "Type of content of item; e.g., "still image", "text", etc."

Sketching an example:

```json
{
  "content_type": "Geospatial data",   # this feels pretty well established
  "format": "Shapefile",
  "resource_type": [
    {
      "value":"Polygon",
      "kind": "Data Type"
    },
    {
      "value":"Vector",
      "kind": "Data Type"
    }
]
}
```

Given the GIS data, it seems that whatever we name this field it should allow multiple values (some MIT and OGM records have 3-4 "Data Type" qualities in the metadata).  I'd worry that `content_type` and `resource_type` seem conceptually similar, but `content_type` is single value string, while `resource_type` is multivalued object.  And, that `format` kind of sits between them hierarchically.

JPrevost: I think `literary_form` could still collapse into this object with either the name `form` or `resourceType`.  I'm not sure which I prefer yet fwiw. I think `form` has a meaning in our current language that is not how we are using it here. I also think that we aren't using it _wrong_ fwiw, but it isn't the first meaning I would intuit when trying to use the data.

JPrevost: Also noting that our current `content_type` appears to be an array of Strings and not just a single String.

JPrevost: Also nothing that our mappings for our sources should be confirmed.

We have thesis records from Alma and DSpace for the same thesis that have different content_types, literary_forms, and formats.

This is not directly related to if we create a new object, but the lack of consistency across sources feels important to note as we consider what these closely related fields _mean_.

Example:
```json
{
  "source": "DSpace@MIT",
  "title": "Gardens of resistance",
  "contributors": [
    {
      "kind": "author",
      "value": "Jhaveri, Nynika\n            (Nynika P.)"
    },
    {
      "kind": "other",
      "value": "Massachusetts Institute of Technology. Department of Architecture."
    },
    {
      "kind": "department",
      "value": "Massachusetts Institute of Technology. Department of Architecture"
    }
  ],
  "contentType": [
    "Thesis"
  ],
  "literaryForm": null,
  "format": "electronic resource"
},
{
  "source": "MIT Alma",
  "title": "Gardens of resistance",
  "contributors": [
    {
      "kind": "Not specified",
      "value": "Jhaveri, Nynika (Nynika P.)"
    },
    {
      "kind": "Not specified",
      "value": "Massachusetts Institute of Technology. Department of Architecture"
    }
  ],
  "contentType": [
    "Manuscript language material"
  ],
  "literaryForm": "Nonfiction",
  "format": null
}
```

ghukill: > Also noting that our current `content_type` appears to be an array of Strings and not just a single String.

Good catch!  Thanks.  Crossing that out from above.

docs/adrs/0002-field-for-data-type-form-information.md
```diff
@@ -0,0 +1,162 @@
...
+### Overview
+During a project that introduced geospatial data into TIMDEX, two new sources were added `gisogm` and `gismit`, both of which use the new Transmogrifier transformer 
+`MITAardvark`. Records for these sources come from the [GeoHarvester](https://github.com/MITLibraries/geo-harvester) which produces [Aardvark metadata schema](https://opengeometadata.org/ogm-aardvark/) JSON records.    
+
+During work on a new geospatial TIMDEX UI, the question was raised as to what fields should be used for two new required facet filters to mirror functionality in the [current Geoweb application](https://geodata.mit.edu/):
+
+  * "Data Type"
+    * values like: `[Polygon, Point, Raster, Image, etc.]`
+  * "Format"
+    * values like: `[Shapefile, TIFF, GeoTIFF, JPEG, etc.]`
```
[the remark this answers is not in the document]
JPrevost: Hmmm. I think that might be another option to document?

i.e. move `[Shapefile, TIFF, GeoTIFF, JPEG, etc.]` to file_formats... and `[Polygon, Point, Raster, Image, etc.]` slide into `format`... no data model changes required?

(I'm not saying I'd pick that option, but it feels like an option to document)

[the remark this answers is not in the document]
ehanson8: I can write that up and push a commit, if that works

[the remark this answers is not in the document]
ghukill: Thanks for suggesting @ehanson8.  `file_formats` felt like a non-starter because other Transmog sources were writing mimetypes... but would probably be relatively easy to update them.

I'm liking that proposal quite a bit, but did want to note that would mean the "Format" filter in the UI would actually be coming from `file_formats` and "Data Type" filter would be coming from `format`.   Would probably want to document that in the UI why this mapping was selected.

For any other sources that write to `format`, I'm assuming we might want to consider if the values they are writing might then belong in `file_formats`?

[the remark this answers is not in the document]
ehanson8: Pushed commit with the 4th option

docs/adrs/0002-field-for-data-type-form-information.md
```diff
@@ -0,0 +1,184 @@
...
+}
+```
+
+Pros:
+  * does not require TIMDEX data model changes
+
+Cons:
+  * `file_formats` has previously only stored MIME type values, such as `application/pdf`
+  * may require explanation of the facet mapping in the UI documentation
+  * may require updates of other transform classes for consistency
```
ghukill, line 147: I'm finding Option #4 pretty appealing now.  

One question: what if we had a record for a print map?  If the Aardvark record had `dct_format_s="Print map"`, are we okay with this value mapping to `file_formats`?  

Maybe helpful for discussion, [this spreadsheet](https://docs.google.com/spreadsheets/d/10oNtwD-3HagVHFfa00sB8eWs5EVmmWMiLSq2q08VU-w/edit#gid=1635328382) has `dct_format_s` and `gbl_resourceType_sm` values across all MIT and OGM records.

In support of Option #4, I'm hard pressed to find a value that _isn't_ a digital file format in essence (where even msot of the complex ones could be mapped to "Online resource" or something similar).

ghukill: Another question: if we go this route, are we okay losing mimetype as a value in a TIMDEX record?  My inclination is that it would be okay as I'd imagine this field is not widely used yet.

If we did want specifics about mimetypes, perhaps that could get added to the `links` section?  Seems more meaningful for a mimetype to point _precisely_ to an actual resource vs the record as an abstract whole.

ehanson8: > If we did want specifics about mimetypes, perhaps that could get added to the `links` section? Seems more meaningful for a mimetype to point _precisely_ to an actual resource vs the record as an abstract whole.

Agree that including it with the actual download link would be more useful

ehanson8: > One question: what if we had a record for a print map? If the Aardvark record had `dct_format_s="Print map"`, are we okay with this value mapping to `file_formats`?

If we do go this route, we may want to change the name from `file_formats` so it's more broadly applicable but I don't have a suggestion yet

ghukill: > > One question: what if we had a record for a print map? If the Aardvark record had `dct_format_s="Print map"`, are we okay with this value mapping to `file_formats`?
> 
> If we do go this route, we may want to change the name from `file_formats` so it's more broadly applicable but I don't have a suggestion yet

I've had time to look more closely at the spreadsheet linked above, and I don't think we'll encounter "Print Maps", at least in the `gismit` or `gisogm` sources.  So maybe this is moot?  I think it's safe to say that the values we get back from `dct_format_s` in the MITAardvark files will uniformly be digital in nature.  And therefore, `file_formats` could be applicable.

ghukill: > > If we did want specifics about mimetypes, perhaps that could get added to the `links` section? Seems more meaningful for a mimetype to point _precisely_ to an actual resource vs the record as an abstract whole.
> 
> Agree that including it with the actual download link would be more useful, `Link.file_type` attribute?

If we did repurpose `file_formats`, and therefore needed to move those mimetype data points to another field, and we _did_ like putting them on the `links` entries, I might propose `mimetype` explicitly; no ambiguity there.

If we go that route, I think we'd want to make sure the links made sense though.  If we're just pointing someone back to an Alma record, or ASpace page, I'm not sure that an `link.mimetype=application/pdf` makes sense.

jonavellecuerdo:
@ghukill and @ehanson8 Thank you for writing up detailed descriptions for possible solutions. Here are my thoughts:

* **Option 1**: Didn't feel strongly; he reason stated in the "Cons".
* **Option 2**: Didn't feel strongly; I did think the term "Form" was a bit confusing. 
* **Option 3**: Kinda' dug this solution; **But** I still thought "Form" didn't suit "data types" or "file formats". 
* **Option 4**: Definitely felt like this was the most understandable solution. ✅ 

Question: Would it be a big lift if we were to:

* Create a new class `transmogrifier.models.FileFormatType`:
   ```
   class FileFormatType:
      value: str = field(validator=instance_of(str))  # Required subfield
      kind: str | None = field(default=None, validator=optional(instance_of(str)))
   ```
* Update TIMDEX model `TimdexRecord`
   ```
   file_formats: list[str] | None = field(default=None, validator=optional(list_of(FileFormatType)))
   ```

Such that....we can specify---

```
FileFormatType(value="application/pdf", kind="MIME")
FileFormatType(value="Shapefile", kind="Image file format")
```

ghukill:
As if this discussion doesn't have enough context, wanted to add a bit more.  

After some normalization of terms in the GeoHarvester (specifically `dct_format_s` to [these terms](https://opengeometadata.org/ogm-aardvark/#format-values), and `gbl_resourceType_sm` to [these](https://opengeometadata.org/ogm-aardvark/#resource-type-values-loc) and [these](https://opengeometadata.org/ogm-aardvark/#resource-type-values-ogm) terms), here are the counts across MIT and OGM records: https://docs.google.com/spreadsheets/d/10oNtwD-3HagVHFfa00sB8eWs5EVmmWMiLSq2q08VU-w/edit#gid=61712783.

This gives a pretty good preview of what kind of actual values we're looking to slot into the data model, and see in the UI.

**One observation**: for the OGM `gbl_resourceType_sm` field, you can see a longer tail of values than the MIT records.  This is because the Aardvark schema allows for either LOC or OGM terms here.  

I would propose we filter this down in Transmogrifier to only allow the values we want to see for "Data Types", e.g. "Polygon", "Point", etc., and then put the rest into `subjects`, as they are quite nicely formed and we even have some namespace/provenance to understand where they come from.

jonavellecuerdo:
> > @ghukill and @ehanson8 Thank you for writing up detailed descriptions for possible solutions. Here are my thoughts:
> > 
> > * **Option 1**: Didn't feel strongly; he reason stated in the "Cons".
> > * **Option 2**: Didn't feel strongly; I did think the term "Form" was a bit confusing.
> > * **Option 3**: Kinda' dug this solution; **But** I still thought "Form" didn't suit "data types" or "file formats".
> > * **Option 4**: Definitely felt like this was the most understandable solution. ✅
> > 
> > Question: Would it be a big lift if we were to:
> > 
> > * Create a new class `transmogrifier.models.FileFormatType`:
> >   ```
> >   class FileFormatType:
> >      value: str = field(validator=instance_of(str))  # Required subfield
> >      kind: str | None = field(default=None, validator=optional(instance_of(str)))
> >   ```
> > * Update TIMDEX model `TimdexRecord`
> >   ```
> >   file_formats: list[str] | None = field(default=None, validator=optional(list_of(FileFormatType)))
> >   ```
> > 
> > Such that....we can specify---
> > ```
> > FileFormatType(value="application/pdf", kind="MIME")
> > FileFormatType(value="Shapefile", kind="Image file format")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> A couple thoughts here...
> 
> If we did this, I'd propose naming the class `class FileFormat` which matches the `CamelCase` style of other field-specific classes that slow into a `snake_case` form in the final record.
> 
> As for this change itself, a couple of considerations:
> 
> * for any sources that currently write `file_formats`, we'd have to update transformation and reharvest to ensure they now have a complex object with `value/kind`
> * I think we might struggle with programatically applying `kind` to non-mimetype values, where it's a bit subjective how to classify them.
> 
> Having been down this road a bit with preservation systems, I think we might want to stay out of the game of trying to juggle "friendly" and mimetype forms. There are python library we could lean on to give us one from the other, but there are _always_ edge cases.
> 
> Zooming out a bit -- something I think Option 4 kind of raises as a question -- perhaps we don't need mimetypes at all? Or, not needed for `file_formats`? And/or, add `mimetype` to `links` _only_ as floated above?
> 
> To me, the appeal of Option 4 would be:
> 
> * no structural changes to TIMDEX data model; `file_formats` remains a list of strings
> * update sources that do currently write to `file_formats` to write "human" friendly forms, e.g. `"PDF"` instead of `application/pdf` which would then match the "human" friendly forms that Aardvark would introduce like `GeoTIFF` or `Shapefile`.

Given the considerations you mentioned, it seems like the additional changes I proposed would prove challenging, so please feel ignore! I do like the idea of "updat[ing] sources that do currently write to file_formats to write "human" friendly forms". Curious to see what others think! 🤓

matt-bernhardt (commented):
I have some concerns with the third option, which I've tried to explain above. Ultimately, though, I don't know that I'm familiar enough with this problem space to attempt to veto any of the four options.

I have a separate question about our change processes which this has brought up for me - how significant an undertaking is it for the data model to change? Several of these options include "pro" statements about not requiring the data model to change, and I want to make sure I understand that prospect accurately. Are we avoiding changes to the data model because:

* Doing so actively causes problems, or 
* Because it is a complex process that must be undertaken thoughtfully (and which takes significant time), or
* Because we've yet to make these sorts of data model changes, so the process is completely undefined?

jazairi (approved):
I support this decision. While losing `Geospatial data` is a shame, I agree that considering `content_class` is out of scope of this ADR. (Though, I think I'd support adding that field, too.) In the meantime, if we need to identify whether a record is geospatial, we should be able to infer as much from `source`.

Thank you to the DataEng team for your thoughtful consideration of this issue, and to Graham for the clear and detailed write-up!

jonavellecuerdo (approved):

ehanson8 (approved):
I'm also in favor of option 5, the added examples were very helpful

matt-bernhardt (commented):
Thanks for facilitating this conversation, everyone. Based on the way the document now reads, I'd support choosing option 5 from those laid out. I'm also likely in favor of the future addition of a `content_class` field.

The example values across a variety of sources was particularly helpful in coming to terms with this.

I see that the Decision area is still left as TBD, so I'm not sure that I'm in a position to approve as-is - but I think I'm with the others so far who have said they support that choice.

docs/adrs/0002-field-for-data-type-form-information.md
```diff
@@ -0,0 +1,253 @@
+# 2. Data Model Fields to Support "Format" and "Data Type" UI Filters
+
+Follows: https://mitlibraries.github.io/guides/misc/adr.html
+
+Date: 2024-02-15
+
+## Status
+
+Proposed
```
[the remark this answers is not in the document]
JPrevost: Sorry, ADR practice is actually `Accepted` for this status, not `Approved`

ghukill:
> I'd like my approval to signal a specific option. Could you update the status to `Approved` and the decision to `Option 5` and dismiss all current reviews.
> 
> I do support option 5, but it's unclear if all of the other approvals are approving option 5 or something else because the document doesn't have a decision listed :)
> 
> Thanks!

Thanks for the ADR logistical help here @JPrevost!  ADR marked as "Accepted", decision logged, and all reviewers have been re-requested.

jazairi (approved):

ehanson8 (approved):

JPrevost (approved):

jonavellecuerdo (approved):

PR #132, opened 2024-02-23
Adds ADR for remapping PublicationInformation

JPrevost:
also adds config to allow adr cli to create an adr in the same directory other adrs are found in this repository

https://mitlibraries.atlassian.net/browse/GDT-199

Rendered markdown of current ADR: https://github.com/MITLibraries/transmogrifier/blob/gdt-199-publication-adr/docs/adrs/0003-support-aggregations-on-publisher-name.md

NO

YES

### DIFF
--- a/docs/adrs/0003-support-aggregations-on-publisher-name.md
+++ b/docs/adrs/0003-support-aggregations-on-publisher-name.md
@@ -88,6 +88,84 @@ GraphQL would continue to serve `PublicationInformation`, but replace it with th
 | Alma | [MARC 260$c](https://www.loc.gov/marc/bibliographic/bd260.html) | Dates.kind=PublicationDate |  (this may already happen?) |
 | Alma | [MARC 264$c](https://www.loc.gov/marc/bibliographic/bd264.html) | Dates.kind=PublicationDate |  (this may already happen?) |
 
+### Proposed mappings (Option 3)
+
+This option is a combination of Option 1 and 2.
+
+In this scenario:
+- **From Option 1**: all sources write publisher information to a multivalued object field `PublicationInfo` with fields like `[name, date, location, etc.]` 
+  - there is no normalization or parsing of the data; strings are written as found from the original record
+- **From Option 2**: where data is available (most commonly with Alma) sources extract date and location from the publisher information and write those values to `Dates` and `Locations` respectively, with a `@kind=Publisher` qualifier
+  - in the case of dates, we _could_ normalize and validate the date string to ensure it's a valid and meaningful Opensearch date
+
+Advantages of this option:
+- all information for a specific publisher (e.g. name, date, location) is contextualized together as a complex object under `PublicationInfo`
+  - e.g. we can know "the published date via the 'Great Writings' publisher is 1930" 
+- for TIMDEX UI search and item pages, and GraphQL aggregations, there is no need to dig into complex objects
+  - simply look to `Dates` or `Locations` for that information where this data has been duplicated
+- logic for extracting dates and locations from publisher information could be shared across all Transmogrifier sources
+  - e.g. it could be an automatically applied, secondary step after the `PublicationInfo` objects are created, pulling from `PublicationInfo.date` and `Publication.location`
+  - allows for more thorough date parsing for `Dates` entries, without losing meaningful strings from the source record that can remain in the `PublicationInfo` object
+
+Example record:
+```json
+{
+  "publication_info": [
+    {
+      "name": "Great Writings",
+      "date": "1930",
+      "location": "Bend, OR"
+    },
+    {
+      "name": "Amazon Reprints",
+      "date": "2020",
+      "location": "Seattle, WA"
+    },
+    {
+      "name": "Ebooks Inc.",
+      "date": "Circa 2023?"
+    }
+  ],
+  "dates": [
+    {
+      "kind": "Published",
+      "value": "1930"
+    },
+    {
+      "kind": "Published",
+      "value": "2020"
+    }
+  ],
+  "locations": [
+    {
+      "kind": "Published",
+      "value": "Bend, OR"
+    },
+    {
+      "kind": "Published",
+      "value": "Seattle, WA"
+    }
+  ]
+}
+```
+- note that bad or missing data from publisher "Ebooks Inc." is skipped for `dates` and `locations` extraction
+
+This avoids some subtle but potentially confusing scenarios:
+
+- **Option 1**: a user clicks date facet "1910" in search UI but does not see "1910" under "Dates" in the item page
+  - **Reason**: the UI item page didn't know it should reach into `PublicationInfo` objects for dates to show under "Dates", as this was custom logic applied to GraphQL aggregations and search facets
+  - **Option 3 fix**: GraphQL, UI search, and UI item pages all pull publishers from publishers, dates from dates, locations from locations, etc., no logic required
+
+- **Option 2**: a user is viewing an item page for the geospatial record "Fires in 1999 Dataset" but sees a strange "2020" under the "Dates" section
+  - **Reason**: the item page "Publisher" section only shows "GIS Pro Inc.", as "2020" was decoupled from the publisher (`Date` object would contain `@kind=Published` qualifier, but we'd need to then include that in the item page)
+  - **Option 3 fix**: the "Publisher" section in the item page clearly also shows "2020", because still part of a complex object, thereby contextualizing that date
+
+Option 3 achieves data where and when needed, and with the appropriate amount of context, by _extracting and duplicating_ some data like dates and locations:  
+
+- a user wants details about a publisher, look at full and complex `PublicationInfo` object in the record
+- the API or UI wants to pull all meaningful dates or locations from a record, look to the `Dates` or `Locations` fields
+
+In either situation, no additional logic, mapping, or documentation is needed. 
 
 ## Decision
 
--- a/docs/adrs/0003-support-aggregations-on-publisher-name.md
+++ b/docs/adrs/0003-support-aggregations-on-publisher-name.md
@@ -38,11 +38,13 @@ Create a new object `PublicationInfo` and map data into it. Most sources just us
 
 ```json
 {
-  publicationInfo: [{
-    name: String
-    location: String
-    date: String (ideally dates would be dates... but that may be out of scope for now)
-  }]
+  "publicationInfo": [
+    {
+      "name": "String",
+      "location": "String",
+      "date": "String (ideally dates would be dates... but that may be out of scope for now)"
+    }
+  ]
 }
 ```
 
@@ -68,9 +70,9 @@ GraphQL would collapse the new object and continue to serve it as the deprecated
 
 ### Proposed mappings (Option 2)
 
-Move some of the Alma data to other existing objects, and create a new top level `Publisher` array of string.
+Move some of the Alma data to other existing objects, and create a new top level `Publisher` array of strings.
 
-All sources except Alma remap to use `PublicationInfo` instead of `PublicationInformation` with no other changes at this time.
+All sources except Alma remap to use `Publisher` instead of `PublicationInformation` with no other changes at this time.
 
 This decouples our data model more from MARC where we seem to have modeled a set of data to match 260/264 in a way that doesn't seem to map to our other sources. Additionally, we already have other places for the extra info from 260/264 to map that may be more useful than as a new Object or embedded into the original String.
 
--- a/docs/adrs/0003-support-aggregations-on-publisher-name.md
+++ b/docs/adrs/0003-support-aggregations-on-publisher-name.md
@@ -171,7 +171,12 @@ In either situation, no additional logic, mapping, or documentation is needed.
 
 ## Decision
 
-Coming soon
+Proceed with Option 3:
+
+- create new, top level, multivalued object field `PublicationInfo` with properties `[name, date, location]`
+- where possible, further parse dates and locations from `PublicationInfo` objects into `Dates` and `Locations` fields, with `@kind=Publisher` qualifier
+- all pre-existing transformations begin writing to `PublicationInfo` instead of current multivalued string `publication_information`
+- deprecate `publication_information` in GraphQL, point to new object field `PublicationInfo`
 
 ## Consequences
 
--- a/docs/adrs/0003-support-aggregations-on-publisher-name.md
+++ b/docs/adrs/0003-support-aggregations-on-publisher-name.md
@@ -70,9 +70,11 @@ GraphQL would collapse the new object and continue to serve it as the deprecated
 
 ### Proposed mappings (Option 2)
 
-Move some of the Alma data to other existing objects, and create a new top level `Publisher` array of strings.
+Create a new top level `Publisher` array of strings.
 
-All sources except Alma remap to use `Publisher` instead of `PublicationInformation` with no other changes at this time.
+All sources (including Alma) would begin to write the publisher name to the new multivalued string `Publisher` field.
+
+Additionally, any sources (most commonly Alma) that have publisher date or location information, could extract and write that data to other appropriate fields, e.g. `Dates` or `Locations` with a qualifier like `@kind="Published"`.
 
 This decouples our data model more from MARC where we seem to have modeled a set of data to match 260/264 in a way that doesn't seem to map to our other sources. Additionally, we already have other places for the extra info from 260/264 to map that may be more useful than as a new Object or embedded into the original String.
 
--- a/docs/adrs/0003-support-aggregations-on-publisher-name.md
+++ b/docs/adrs/0003-support-aggregations-on-publisher-name.md
@@ -1,0 +1,96 @@
+# 3. Support Aggregations on Publisher Name
+
+Date: 2024-02-23
+
+## Status
+
+Accepted
+
+## Context
+
+Our current data model maps data about Publisher Name, Publication Year, and Publication Location into a single multivalued field (an Array of Strings). Mapping these different concepts into a single field makes the data less meaningful and less useful than it could be with a change to our data model.
+
+Mapping differnt types of data into a single field makes aggregation confusing as we'd see things like "1999" along side "Massachusetts Institute of Technology" which is not usable.
+
+### Current mappings
+
+| Source | Mappings | TIMDEX Field | Examples |
+| --- | --- | --- | --- |
+| Alma | [MARC 260$abcdef](https://www.loc.gov/marc/bibliographic/bd260.html) | PublicationInformation | Charlottesville : University of Virginia Press, [2015], ©2015 |
+| | [MARC264$abc](https://www.loc.gov/marc/bibliographic/bd264.html) |  PublicationInformation | DeKalb, Illinois : NIU Press, 2014, ©2014 |
+| |   | PublicationInformation | null |
+| |   | PublicationInformation  | A.O.K. 3 |
+| |   | PublicationInformation  | American Antiquarian Society Historical Periodicals |
+| DSpace@MIT (METS) | XML slect array of Strings matching element `publisher` | PublicationInformation | Massachusetts Institute of Technology |
+| |   | PublicationInformation  | Elsevier BV |
+| |   | PublicationInformation  | null |
+| |   | PublicationInformation | ACM|Creativity and Cognition |
+| |   |  PublicationInformation | Cambridge, Mass. : Alfred P. Sloan School of Management, Massachusetts Institute of Technology |
+| |   | PublicationInformation  | Elsevier |
+| |   | PublicationInformation  | Wiley |
+| ArchivesSpace | XML select array of Strings matching element `dc:publisher`  | PublicationInformation  | Massachusetts Institute of Technology. Libraries. Department of Distinctive Collections (note: this appears to be effectively a static string for all of our ASpace records) |
+| MIT GIS |   | PublicationInformation |  |
+| OGM GIS |   | PublicationInformation  |  |
+
+### Proposed mappings (Option 1)
+
+Create a new object `PublicationInfo` and map data into it. Most sources just use `Publication.name` but Alma uses additional fields.
+
+```json
+{
+  publicationInfo: [{
+    name: String
+    location: String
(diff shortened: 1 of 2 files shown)


docs/adrs/0003-support-aggregations-on-publisher-name.md
```diff
@@ -0,0 +1,96 @@
...
+{
+  publicationInfo: [{
+    name: String
+    location: String
+    date: String (ideally dates would be dates... but that may be out of scope for now)
+  }]
+}
+```
+
+Notes: we should consider whether this really needs to be an array. 260 and 264 are repeatable fields in MARC, but whether they are regularly used in that way or if we could instead pick the "first" one that shows up and get a simpler model without losing any real world functionality would be worth investigating. For now it is modeled as an array of `PublicationInfo` assuming that is appropriate.
```
ghukill, line 49: From EngX's / API POV, what makes a single value simpler than an array?  Is it parsing those results from the actual document in GraphQL?

ghukill:
At this time, I'm leaning towards Option 2 for a couple of reasons.

First, I found this quite compelling:
> This decouples our data model more from MARC where we seem to have modeled a set of data to match 260/264 in a way that doesn't seem to map to our other sources

Second, when you extend that thinking further, though I'm confident we _will_ encounter publisher dates, it's more likely that we'd want those dates when already in the context (i.e. facet filter) of thinking about dates.  This, as opposed to rooting around in an object for them.

This might suggest a somewhat general pattern where _some_ fields in TIMDEX benefit from having `kind` or other kinds of qualifiers, e.g. Dates and Locations.  We may have date and location information for other descriptive components -- like Publisher -- but we often don't need to know all that at once.  Storing a `date@kind=publisher` and `location@kind=publisher`, as long as the `kind` values are sensible and controlled (which we could enforce in Transmog), makes that information more where you need and expect it.

docs/adrs/0003-support-aggregations-on-publisher-name.md
JPrevost: It's less EngX than end users I'm thinking about. Nested objects can be more difficult to work with in GraphQL than the equivalent top level field. And strings are easier than arrays to work with initially. This doesn't mean we should aim for totally flat records of entirely strings, just that when our data does not require objects or arrays, we should lean towards strings.

jazairi: It seems like we should make it an array given that it's possible for multiple occurrences, but I agree that it would be useful to find out how often that happens, if ever, to inform that decision. It would be kind of silly if we accommodated repeatable MARC fields that are never actually repeated.

docs/adrs/0003-support-aggregations-on-publisher-name.md
```diff
@@ -0,0 +1,98 @@
...
+| Alma | [MARC 264$c](https://www.loc.gov/marc/bibliographic/bd264.html) | PublicationInfo.Date | |
+| Alma | [MARC 260$d](https://www.loc.gov/marc/bibliographic/bd260.html) | Invalid field. Don't map | |
+| Alma | [MARC 260$e](https://www.loc.gov/marc/bibliographic/bd260.html) | Don't map | |
+| Alma | [MARC 260$c](https://www.loc.gov/marc/bibliographic/bd260.html) | Don't map | |
+| DSpace | Keep logic | Publication.Name | note: consider future work to normalize in source or during transform |
+| ASpace | Keep logic | PublicationInfo.Name | |
+| GIS MIT | | PublicationInfo.Name | |
+| GIS OGM | | PublicationInfo.Name | |
+
+### Proposed mappings (Option 2)
```
jonavellecuerdo, line 71: @JPrevost For Option 2, would it also require only choosing the _first_ occurrence of the MARC 260 and 264 fields? 🤔

JPrevost: Good question! I believe similar to the discussion in Option 1, the fields are repeatable but it's unclear how frequently they are actually repeatable. I'd suggest we may want to look at our data to decide if this should be an Array of Strings or a single String. I suspect an Array of Strings will be where we end up but I'd hate to make that decision based on suspicion rather than data.

docs/adrs/0003-support-aggregations-on-publisher-name.md
```diff
@@ -0,0 +1,98 @@
...
+
+Move some of the Alma data to other existing objects, and create a new top level `Publisher` array of strings.
+
+All sources except Alma remap to use `Publisher` instead of `PublicationInformation` with no other changes at this time.
```
ghukill, line 75: Curious about this line:
> All sources except Alma remap to use `Publisher` instead of `PublicationInformation` with no other changes at this time.

I realized on my first read, I had made the assumption:
- all sources (including Alma) would begin to write the publisher *name* to this new multivalued string `publisher` field
- future work: any sources (including Alma) that have publisher date or location information, could write to other appropriate fields, e.g. `dates` or `locations` with a qualifier like `@kind="Publisher"`

Does this align with your thinking and the wording here?

docs/adrs/0003-support-aggregations-on-publisher-name.md
ghukill: Just to tease this out a bit more... if:

- `publisher` is a multivalued string field
- and any other publisher information gets written to other fields like `dates` or `locations` ([see comment below](https://github.com/MITLibraries/transmogrifier/pull/132/files#r1506065717))

wouldn't this mean, for Option 2, that we could accomodate both MARC 260 and 264?  Stated more generally, that Option 2 supports multiple publishers by virtue of the new `publisher` field being multivalued and all other values mapped to other, multivalued fields?

JPrevost: Correct. If we just make `publisher` an array of Strings, we can support any source that currently or in the future has multiple values for all aspects of this (dates, names, locations).

If we make `publisher` a single string, we could still support multiple Dates/Locations if that is the only thing that is truly multivalued in our sources. That may be a bit hard to tease out if that is all we truly need or if we should just accept that supporting multivalued is better because we just aren't sure.

ghukill:
> At this time, I'm leaning towards Option 2 for a couple of reasons.
> 
> First, I found this quite compelling:
> 
> > This decouples our data model more from MARC where we seem to have modeled a set of data to match 260/264 in a way that doesn't seem to map to our other sources
> 
> Second, when you extend that thinking further, though I'm confident we _will_ encounter publisher dates, it's more likely that we'd want those dates when already in the context (i.e. facet filter) of thinking about dates. This, as opposed to rooting around in an object for them.
> 
> This might suggest a somewhat general pattern where _some_ fields in TIMDEX benefit from having `kind` or other kinds of qualifiers, e.g. Dates and Locations. We may have date and location information for other descriptive components -- like Publisher -- but we often don't need to know all that at once. Storing a `date@kind=publisher` and `location@kind=publisher`, as long as the `kind` values are sensible and controlled (which we could enforce in Transmog), makes that information more where you need and expect it.

Further support of Option 2, seems like there is similar thinking in MODS for `<origininfo>` where the publisher + event is recorded, but then associated dates (and could conceptually extend this to locations) are stored in that respective part of the record: https://www.loc.gov/standards/mods/userguide/origininfo.html.

ehanson8 (commented):
Agree with everyone else so far on Option 2, the decoupling from MARC will make the MARC transform logic more complicated but I think it's better in the long run

ghukill:
FWIW, this might suggest that Option 1 (array of complex publisher objects) aligns more closely with MODS: https://www.loc.gov/standards/mods/userguide/origininfo.html.

In the MODS schema, the `<origininfo>` parent element has child elements for place, publisher, date, location, etc. 

To me, the difference between Option 1 and 2 feels like it gets at a bigger data model decision: do we store subjects/dates/locations under complex objects like publishers, contributors, etc.?  or do we store those as simple arrays of strings, and then qualify any subects/dates/locations with a `@kind` qualifier and store them in `subjects`/`dates`/`locations`/etc.?

The decision for this ADR, feels like it would set precedence for that going forward, as I'm not seeing this situation in any other fields (complex objects with child dates/locations).  

I think that I continue to support Option 2, and lean into that model:
- simple, top level arrays of strings where it makes sense
- any subjects / dates / locations get routed to those fields, with meaningful and ideally somewhat controlled `@kind` qualifiers

To me, this is more flexibile in the long run.  As we encounter subjects / dates / locations that should be in the record, but don't fall under a complex object, we can continue to slot them into those fields, but ensure (and document) what the `@kind` qualifiers are.

docs/adrs/0003-support-aggregations-on-publisher-name.md
jonavellecuerdo: @JPrevost Does this change mean `TimdexRecord.publication_information`/`PublicationInformation` will be deprecated? 🤔

jonavellecuerdo:
**Option 2** sounds good to me, especially after looking into `all-current` index on OpenSearch (shared in [this comment](https://github.com/MITLibraries/transmogrifier/pull/132#discussion_r1507657483). Approved on my end once decision is officially written into the ADR!

docs/adrs/0003-support-aggregations-on-publisher-name.md
JPrevost: Awesome, thanks for the peek into the data. This is extremely helpful context. I feel like I have enough info to add some additional context to the Options and propose a Decision (option 2 seems popular with the top level array for publisher names and everything else broken into Location and Dates).

jazairi:
In case any further coalescence is needed, I support Option 2 at this point, if not just for flexibility and ease of use.

docs/adrs/0003-support-aggregations-on-publisher-name.md
matt-bernhardt: What sort of warnings get generated when Transmog encounters multiple values when a single one is expected? I'm trying to understand what would happen if, based on an analysis of our current catalog, we decide that multiple values are not supported - but next month something changes and we start seeing them.

matt-bernhardt (commented):
For those cases where we have multiple publishers (not just multiple publication years), would Option 2 introduce potential confusion about the relationship between years, locations, and names among the publishers? Am I interpreting these options correctly with these arrangements?

## Option 1
```
{
  "publicationInfo": [
    {
      "name": "Massachusetts Institute of Technology",
      "location": "Cambridge, MA",
      "date": "2014"
    },
    {
      "name": "Wiley",
      "location": "Hoboken, NJ",
      "date": "2016"
    }
  ]
}
```

## Option 2
```
{
  "date": [
    {
      "kind": "Publication Date",
      "value": "2016"
    },
    {
      "kind": "Publication Date",
      "value": "2014"
    }
  ],
  "location": [
    {
      "kind": "Publisher Location",
      "value": "Cambridge, MA"
    },
    {
      "kind": "Publisher Location",
      "value": "Hoboken, NJ"
    }
  ],
  "publisher": [
    "Massachusetts Institute of Technology",
    "Wiley"
  ]
}
```

I'm not sure how much of an issue this is - it might be fine.

JPrevost:
@matt-bernhardt great question. Yes, Option 1 provides an ability Option 2 does not that we hadn't stated (and I hadn't noticed). Namely, if there is value in keeping the publisher related date together when there are multiple publishers, Option 1 provides that in a way that Option 2 does not.

Based on how infrequently there appear to be multiple _real_ publishers (versus currently a lot of the extra publishers are actually just years inserted into the array), this may not be a concern. However, it should be noted as a distinct feature difference between the two options for sure and noted in the consequences section if we choose the one that would lose this ability to keep these fields together in an object.

ghukill:
Agreed, that @matt-bernhardt's observation is an important one, namely that we decouple the publisher name from the its dates/locations and lose that tether.

Maybe this is not problematic if we continue to view these TIMDEX records through a discovery lens, and not a canonical record lens.  And, potentially even _more_ accurate when we consider the behavior of facets.

Consider this situation with two publishers:

1. "Great Writings; Bend, OR; 1980"
2. "Amazon Reprints, Seattle, WA; 2020"
3. "Ebooks Inc., Believed to be 2023 but records are hazy"

If this record comes up in a search, and a user clicks the publisher facet for `"Great Writings"`, if we tightly couple the publisher name + location + date, we'd _lose_ the location "Seattle, WA" and the date "2020" in the facets.  However, I'd float those facet values are still relevant to the record, as they are part of its metadata in its totality.

The flipside -- and what I assume is the concern raised above -- is that by decoupling them, we would lose the ability to look directly at this TIMDEX record and know:

1. For the "Great Writings" publisher, the date was 1980 and the location was Bend, OR
2. For the "Amazon Reprints" publisher, the date was 2020 and the location was Seattle, WA

I hesitate to suggest it... but what if we combined Option 1 & 2?

```json
{
  "publishers": [
    {
      "name": "Great Writings",
      "date": "1930",   <----- use original string
      "location": "Bend, OR"
    },
    {
      "name": "Amazon Reprints",
      "date": "2020",   <----- use original string
      "location": "Seattle, WA"
    },
    {
      "name": "Ebooks Inc.",
      "date": "Believed to be 2023 but records are hazy"   <----- use original string
    }
  ],
  "date": [
    {
      "kind": "Published",
      "value": "1980-01-01"   <----- parsed as true dates, with an eye towards this desired state
    },
    {
      "kind": "Published",
      "value": "2020-01-01"   <----- parsed as true dates, with an eye towards this desired state
    }
  ],
  "locations": [
    {
      "kind": "Published",
      "value": "Bend, OR"
    },
    {
      "kind": "Published",
      "value": "Seattle, WA"
    }
  ]
}
```

In this scenario, we'd get:

- accurate facets; no loss of information when one publisher identified
- publisher dates/locations slot into `dates` and `locations` for ease of use there
  - furthermore, they are parsed and normalized according to those fields
- for the messy "Ebooks Inc." entry, we maintain the original strings, but we just skip extracting and normalizing dates from it

There is duplication of information, but we'ev also normalized and filtered out bad values while still retaining the original which may have some meaning.

It also might be less jarring to see this in the context of other fields.  If the pattern is: store complex objects, but attempt to extract meaningful dates / subjects / locations from it and store in those respective fields, then maybe it doesn't seem as unusual?

Lastly, from a Transmogrifier POV, we could focus on writing complex Publisher objects, then have shared methods and normalization logic across all sources that attempt to parse out distinct subjects, dates, locations, etc.

Just food for thought!

ehanson8:
> Maybe this is not problematic if we continue to view these TIMDEX records through a discovery lens, and not a canonical record lens.

I still think discovery is the most appropriate lens to view TIMDEX and combining 1 & 2 feels like it's introducing a lot of complexity for accuracy in a likely rare occurrence of multiple legitimate publishers

ghukill:
More comments forthcoming, but I did want to correct a mistake in my [comment from last week](https://github.com/MITLibraries/transmogrifier/pull/132#issuecomment-1973710461).

Specifically, that when selecting a facet value, we don't "lose" the other facet values from that particular record.  See this example of selecting "Polygon data" as a facet value, but because the document contains other values for "Content Type", they still show up:

Screenshot 2024-03-04 at 10 22 44 AM

More comments coming, but this is an important clarification as neither Option 1 or 2 result in the behavior I had suggested above.

docs/adrs/0003-support-aggregations-on-publisher-name.md
ghukill: @JPrevost - language updated in Option 2, per [your aggrement with a slight rewording](https://github.com/MITLibraries/transmogrifier/pull/132#discussion_r1506082129), per this commit: https://github.com/MITLibraries/transmogrifier/pull/132/commits/1dfcfab19a41d5a2c021658e7acd9186c79afb53.

ghukill:
@JPrevost, @matt-bernhardt, @ehanson8, @jazairi, @jonavellecuerdo - an "Option 3" has been added [per this commit](https://github.com/MITLibraries/transmogrifier/pull/132/commits/99173c3fcceafb7b6daca126d0507c333a7970e4) that sketches a combination of Option 1 and Option 2.

Matt's [observation](https://github.com/MITLibraries/transmogrifier/pull/132#pullrequestreview-1911456231) about Option 2 losing context between publisher names and dates/locations felt like an important one.  Personally, this felt like a deal breaker for Option 2 for me.

However, I feel Option 1 may continue to propagate some pain points we are feeling in the TimdexRecord / Opensearch / GraphQL / Search UI / Item Page UI relationshipos where we are required to dig into complex objects for aggregations.  This alone may not be technically problematic, but it could be brittle in the longer term (see example in Option 3 where "Publisher" or "Date" sections start to drift between the search UI filters and the item page representation).

It's hard to succinctly summarize, and might be beyond the scope of this ADR to some degree, but to me Option 3 represents a pattern that could be applied going forward: **store complex objects to preserve context, but also extract subjects / dates / locations for ease of use and consistency of aggregation and item viewing**.

Perhaps Option 3 is not ultimately what we want, but hoping it can help clarify the strenghts and weaknesses of the current Options 1 and 2.

At this time, my vote would be for Option 3.

jonavellecuerdo:
I like **Option 3**. I think [lines 165-168 of the ADR](https://github.com/MITLibraries/transmogrifier/blob/gdt-199-publication-adr/docs/adrs/0003-support-aggregations-on-publisher-name.md?plain=1#L165-L168) summarize it really well. I feel it benefits the end user (by preventing information loss) and developers (by easing navigation of the data model). 

As a side note, [`Aardvark.dct_issue_s `could be updated to use `kind="Publication date"` (instead of "Issued")](https://github.com/MITLibraries/transmogrifier/blob/0fd8ea882733367606f8c13610b3d000a94beef8/transmogrifier/sources/json/aardvark.py#L222-L225) to follow the `timdex.Date` derivations for other transformers.

ghukill:
> As a side note, [`Aardvark.dct_issue_s `could be updated to use `kind="Publication date"` (instead of "Issued")](https://github.com/MITLibraries/transmogrifier/blob/0fd8ea882733367606f8c13610b3d000a94beef8/transmogrifier/sources/json/aardvark.py#L222-L225) to follow the `timdex.Date` derivations for other transformers.

Agreed @jonavellecuerdo about looking for opportunities to standardize the `kind` values we use.  To the degree that we "extract" data from other parts of a record and store them as Dates / Subjects / Locations, it increases the value in standardizing the `kind` qualifiers we use.

It may be out of scope for implementing the final decision of this particular ADR, but a follow-up ADR at some point around normalizing `kind` values would be a great one.

ehanson8:
I'm fine with Option 3 as well. Standardizing `kind` values is something we've wanted to do but just never got around to it, this would be good to prioritize as we go down this road

jonavellecuerdo (approved):

jazairi (approved):
Option 3 sounds good to me, too. I agree with @ghukill that, from the perspective of the API consumer, this structure is no more complex to navigate than the one presented in Option 2. That it also provides richer and clearer data makes this an easy choice for me.

Thanks, all, for the nuanced and thoughtful discussion on this, leading to another excellent decision. 🙂

docs/adrs/0003-support-aggregations-on-publisher-name.md
```diff
@@ -0,0 +1,183 @@
...
+- a user wants details about a publisher, look at full and complex `PublicationInfo` object in the record
+- the API or UI wants to pull all meaningful dates or locations from a record, look to the `Dates` or `Locations` fields
+
+In either situation, no additional logic, mapping, or documentation is needed. 
+
+## Decision
+
+Proceed with Option 3:
+
+- create new, top level, multivalued object field `PublicationInfo` with properties `[name, date, location]`
```
ehanson8, line 176: Can this be `PublicationInformation`? We aren't abbreviating any of the other `transmogrifier` field names or classes

ehanson8 (approved):
Approved but one requested name change

docs/adrs/0003-support-aggregations-on-publisher-name.md
ghukill: I'm a little fuzzy on GraphQL deprecation, but perhaps that is contributing to this variation on `PublicationInformation` which already existed?

Is it correct to assume @JPrevost, @jazairi, @matt-bernhardt, that if we changed the field type of the **pre-existing** `PublicationInformation` field, that doesn't give us the ability to really deprecate it?  Because from GraphQL's POV the field name didn't change, so no ability to say, "This field is deprecated, use XYZ..."?

If so, @ehanson8 and @jazairi (as thumbs-uppers for this suggestion), what would you think about `Publishers` as the complex object field name?  My assumption is that "Information" or "Info" was originally proposed in the event we don't have a name, where just a date or location feels a little odd as a "Publisher".  But, that's generally only the case for copyright dates, and we could elect to skip a `Publisher` entry if only that copyright date, and instead just write that directly to `Dates`.

All considered, I suppose my vote would be `Publishers` as the top level field for complex objects, thereby matching other field name patterns like `Contributors`.

ehanson8: I'm fine with whatever so long as it's not abbreviated 🙂

matt-bernhardt (approved):
I'm in favor of Option 3 as well. I'm not sure what the proper way in GraphQL is for changing a field type, so I'll defer to @JPrevost or @jazairi on that question.

Thanks, everyone, for working through this problem space together. I greatly appreciate everyone's contributions to this process.

docs/adrs/0003-support-aggregations-on-publisher-name.md
jazairi: Oh, yes, good point about the name conflict. I'd be happy with `Publishers` as an alternative.

ghukill: @ehanson8, @jazairi - field name updated from `PublicationInfo` to `Publishers` in this most commit: https://github.com/MITLibraries/transmogrifier/pull/132/commits/9bb06e567a9195c4cdea92c48d6693a17169cf27

ghukill (approved):
While I originally supported Option 2, the comment from @matt-bernhardt about losing the context of dates and locations with respect to the publisher name was a deal breaker for me.

After much discussion and a new option 3 added, I support this decision as well.

JPrevost (commented):
👍🏻 (I can't approve as I opened the PR, but I support the decision to go with Option 3)

PR #133, opened 2024-02-23
Create ADR to propose new field for host institution

jonavellecuerdo:
This PR introduces a new Architecture Decision Record (ADR) that proposes to add a new `institution` field to the TIMDEX data model.

This discussion was kicked off when seeking to understand what TIMDEX fields to use for displaying the host institution on GeoData's "result" and "full record" pages for non-MIT GIS resources.

NO

YES

* https://mitlibraries.atlassian.net/browse/GDT-205

### DIFF
--- a/docs/adrs/0004-field-for-institution-information.md
+++ b/docs/adrs/0004-field-for-institution-information.md
@@ -1,0 +1,125 @@
+# 4. Field for Institution Information
+
+Follows: https://mitlibraries.github.io/guides/misc/adr.html
+
+Date: 2024-02-22
+
+## Status
+
+Approved.
+
+## Context
+
+The UI for GeoData needs to display the host institution of non-MIT records (e.g., "Stanford", "University of Wisconsin-Madison") to signal to end users where an outlink is going. The question was raised of _where_ this information can be found in the existing TIMDEX data model or, if not present, what field will be used to store this information.
+
+### Current mappings
+
+| Source | Mappings | TIMDEX Field | Examples       |
+| ------ | -------- | ------------ | -------------- |
+| Alma   | [MARC$985$i](https://www.oclc.org/bibformats/en/9xx.html) | Holdings.Location | Hayden Library |
+| MIT GIS | [schema_provider_s](https://opengeometadata.org/ogm-aardvark/#provider)| PublicationInformation | GIS Lab, MIT Libraries |
+| OGM GIS | [schema_provider_s](https://opengeometadata.org/ogm-aardvark/#provider)| PublicationInformation | University of Wisconsin-Milwaukee |
+
+1. **Does `geo-harvester` collect the required information (i.e., host institution)?**
+
+   **Yes.** The Aardvark metadata schema includes a field called `schema_provider_s`, which is a single string value describing the [“organization [that] holds the resource or acts as the custodian for the metadata record and to help users understand which resources they can access”](https://opengeometadata.org/ogm-aardvark/#provider). 
+
+   The derivation in `geo-harvester` sets `schema_provider_s` to "GIS Lab, MIT Libraries" for MIT records and the assigned name for the repo in the [OGM repositories config YAML file](https://github.com/MITLibraries/geo-harvester/blob/main/harvester/ogm_repositories_config.yaml).
+
+2. **Does the `MITAardvark` transformer map the required information to a field in the TIMDEX data model?**
+
+   **Yes, but its current mapping does not make the value easily accessible.** In `transmogrifier`, [the value from `schema_provider_s` is appended to `TimdexRecord.publication_information`](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/json/aardvark.py#L382-L383), which comprises of a list of strings describing the publishing details for a given record. The contents of this field will vary by source and depending on what data is available. Trying to derive which value--from a list of strings (without any sort of identifier)--came from `schema_provider_s` is a nontrivial task.
+
+3. **Do the other non-GIS TIMDEX sources have the required information? Are they currently mapped to existing TIMDEX fields?**
+
+   Schemas and sample data for the other sources were explored to find other fields that may identifiy the "host institution" for a given resource:
+
+   ---
+   
+   **Transformer:** [Datacite](https://schema.datacite.org/meta/kernel-4/metadata.xsd) | Zenodo
+   **Source(s):** Abdul Latif Jameel Poverty Action Lab Dataverse (jpal) | Zenodo
+
+   * `<publisher>`: The name of the entity that holds, archives, publishes prints, distributes, releases, issues, or produces the resource. This property will be used to formulate the citation, so consider the prominence of the role. **In the case of datasets, "publish" is understood to mean making the data available to the community of researchers**.
+      * **TIMDEX field mapping**: [`publisher` -> `publication_information`](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/xml/datacite.py#L249-L256)
+      * **Example value(s):** As of this writing, the TIMDEX records from JPAL most likely all read "Harvard Dataverse" for this field; records from Zenodo most likely all read "Zenodo".
+
+   **Transformer:** DspaceDim
+   **Source(s):** Woodshole Open Access Server (whoas)
+   * [`<dim:field mdschema="dc" element="publisher"`>](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/publisher): An entity responsible for making the resource available.
+      * **TIMDEX field mapping:** [`publisher` -> `publication_information`](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/xml/dspace_dim.py#L204-L207)
+      * **Example value(s):** As of this writing, the TIMDEX records from WHOAS most likely all read "Woods Hole Oceanographic Institution" for this field.
+
+   **Transformer:** DspaceMets (DSpace@MIT)
+   **Source(s):** Dspace@MIT (dspace)
+   * [`mods:publisher`](https://www.loc.gov/standards/mods/userguide/origininfo.html#publisher): The name of the entity that published, printed, distributed, released, issued, or produced the resource.
+      * **TIMDEX field mapping:** [`publisher` -> `publication_information`](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/xml/dspace_mets.py#L150-L152)
+      * **Example value(s):** Massachusetts Institute of Technology, Springer Science and Business Media LLC, Public Library of Science (PLoS)
+
+   
+   **Transformer:** [Ead](https://www.loc.gov/ead/tglib/appendix_d.html)
+   **Source(s):** MIT ArchivesSpace (aspace)
+   * [`<publicationstmt><publisher>`|`<bibliography><bibref><imprint><publisher>`](https://www.loc.gov/ead/tglib/elements/publisher.html): When used in the publication statement, the name of the party responsible for issuing or distributing the encoded finding aid. Often this party is the same corporate body identified in the `<repository>` element in the finding aid. When used in a Bibliographic Reference <bibref>, the name of the party issuing a monograph or other bibliographic work cited in the finding aid.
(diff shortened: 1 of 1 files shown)


ghukill (approved):
With these flurry of ADRs and data model discussions, I found @JPrevost's comments helpful in thinking about data model solutions and change over time.  Specifically, how we might have an `institution` field at the GraphQL layer, and while it may point to `TIMDEX.institution` for now, if we choose the move where that data is stored that's just a mapping change at the GraphQL layer.  It gets more complicated when you introduce objects and filtering by sub-fields.

For this reason, I approve this ADR.  Creating a top level field called `institution` is simple and meets the current need.  Thanks for the write-up @jonavellecuerdo!

jazairi (approved):
This feels like the more straightforward solution and the easiest to work with on the UI side. Thanks for the write-up, Jonavelle!

ehanson8 (approved):
I agree with Graham and Adam, good writeup!

docs/adrs/0004-field-for-institution-information.md
```diff
@@ -0,0 +1,68 @@
...
+* `Holding.institution`: This would be a new field, comprised of a single string value, that is added to the `Holding` data model.
+   * There may be some resources that are aggregated by multiple institutions. The concept of "data provider" vs. "provider" was discussed. 
+        * **Provider**: defined as "the institution / organization that hosts or provides access to the resource"
+        * **Data provider**: the institution / organization that provided the metadata record you see before you
+
+## Consequences
+
+For GIS records, the UI can directly reference the values from `TimdexRecord.institution` to display the host institution for a resource on [Geodata](https://geodata.libraries.mit.edu/).
+
+For other TIMDEX sources, given the decision outlined above, there should be little to no effect. It could be beneficial to revisit these sources and see if there is a field we can map to that was previously ignored.
```
JPrevost, line 68: I'm slightly nervous that we don't know yet of other sources will use this field. I'm okay if this is only a useful field for GIS at this time, but the nervousness comes from not knowing if we want to move other things here that indeed make the label of it less good (i.e. `Institution` is perfect for GIS, but `metadata_provider` would work for both GIS and additional sources if we considered mappings more broadly before adding this field. Or not!).

I'd really prefer if we could do the investigation to be able to take a stronger stance here. Either "we don't think this field will be used by any of our current sources so there is no risk" or "we think we'd map fields X from MARC and Y from DSpace and they work perfectly with this plan".

docs/adrs/0004-field-for-institution-information.md
```diff
@@ -0,0 +1,68 @@
+# 4. Field for Institution Information
+
+Follows: https://mitlibraries.github.io/guides/misc/adr.html
+
+Date: 2024-02-22
+
+## Status
+
+Proposed.
```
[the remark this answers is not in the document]
jonavellecuerdo: I will keep this in mind! Thanks for the note. :)

docs/adrs/0004-field-for-institution-information.md
ghukill: Thanks for articulating this @JPrevost, I agree.

Reading this ADR, it felt like a top level, single string field was flexibile enough that we could both modify how it's stored in the data model, and how it's retrieved in the API layer, as needed.  But if it's immediate goal is to support the GIS data, as a thought experiment, why don't we call it `gis_data_resource_provider`?  Obviously, that's too specific; so what then _do_ we mean by `institution`?  are we expecting other sources to potentially use this?

Though I already approved, I think the ADR would benefit from discussion about what `institution` means, and how other sources may, or may not, use it.  Going to leave some inline comments, now that I reread it through this lens.

docs/adrs/0004-field-for-institution-information.md
```diff
@@ -0,0 +1,68 @@
...
+
+   **Yes, but its current mapping does not make the value easily accessible.** In `transmogrifier`, [the value from `schema_provider_s` is appended to `TimdexRecord.publication_information`](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/json/aardvark.py#L382-L383), which comprises of a list of strings describing the publishing details for a given record. The contents of this field will vary by source and depending on what data is available. Trying to derive which value--from a list of strings (without any sort of identifier)--came from `schema_provider_s` is a nontrivial task.
+
+## Decision
+
+**A new field called "institution" is added to the TIMDEX data model** (i.e., `TimdexRecord.institution`). This field will denote--as its name implies--the institution or organization that provides access to the resource described in the TIMDEX record. This field exists at the top level of the TIMDEX data model, making it easily accessible for referencing in the UI or querying. This solution also avoids using existing fields (discussed below) in ways that obscure the TIMDEX data model.
```
ghukill, line 35: On a closer reread, I do think this does a good  job at proposing what this field "means":

> "This field will denote--as its name implies--the institution or organization that provides access to the resource described in the TIMDEX record"

But, if we are thinking of the reusability or naming of this new field, perhaps that "access" word is important.  Is it possible this is really about access primarily?  If so, what if the field were something along the lines of `access_provider`?  Or even just `provider`?

JPrevost: I think `provider` is an interesting concept to consider here. For GIS it seems to map clearly because the provider of the metadata and provider of the resource are generally the same from the TIMDEX perspective.

It seems to get messier for subscription resources that might appear in Alma (which might either help us refine the intent of this field... even if our finding is that this field is not to be used for vended resources :) ). Would the `provider` be the vendor that supplies the content or the subscriber to the content on behalf of the user. i.e. MIT subscribes to an Ebsco database so our users can access the content. Is MIT the `provider` or is `ebsco` the provider. As a consumer service, TIMDEX API should likely take the user perspective?

ghukill: As a consumer service, and a _discovery_ one at that, I suppose I'd expect `provider` to indicate what instituion/organization/company is providing physical or digital access to the resource.

> ...i.e. MIT subscribes to an Ebsco database so our users can access the content. Is MIT the provider or is ebsco the provider.

I like this example because I think a) `provider="Ebsco"` would be ideal, but b) I think that level of provider granularity may vary considerably from source-to-source.  Examples:

- Libguides / Research Database
  - I've confirmed that OAI-PMH records offer no indication of the "leaf" node of access
  - Research Databases: might consider a default of `provider="MIT"`
  - LibGuides: we could say `provider="Springshare"`
- DSpace
  - feels like we could default to `provider="MIT"` for all records
- Alma
  - probably the most complex, where records may have varying degrees of granularity here
- ArchivesSpace
  - perhaps another good example where `provider="MIT"` default makes sense

Don't want to dig too far into the logic, as `provider` is just a consideration at the moment.  But unless we explicitly know -- from a consumer / discovery POV -- maybe a default of `provider="MIT"` wouldn't be inaccurate.

JPrevost: I may lean more towards `provider=unknown` for things like Alma/LibGuides that we aren't sure on... and then take that unknown data to the metadata team and ask if they could make that clearer in the records or how we could better detect it/supplement records from non-alma sources, etc. The reason I'd suggest not defaulting to "MIT" is that would make it very difficult to understand which are actually MIT and which are :shrug:

ghukill: Good point, makes sense.  And maybe to get a bit technical, we could do `provider=NULL`?  Understanding that the absence of a field value means we don't know for certain?  

And clarification: when you say "LibGuides" above, do you mean "ResearchDatabases" (AZ list)?  As for "LibGuides" (they are distinct sources in TIMDEX), it feels like we _do_ know that the provider is always Springshare.

JPrevost: Oh sorry, I read misread and thus misrepresented. LibGuides is probably either MITL or Springshare. The content is curated/created by MITL but using the Springshare platform.

Research Databases is what I meant to be replying to with the "unknown" or your follow on idea of just not setting it (i.e. null).

docs/adrs/0004-field-for-institution-information.md
```diff
@@ -0,0 +1,119 @@
...
+   * [`<publicationstmt><publisher>`|`<bibliography><bibref><imprint><publisher>`](https://www.loc.gov/ead/tglib/elements/publisher.html): When used in the publication statement, the name of the party responsible for issuing or distributing the encoded finding aid. Often this party is the same corporate body identified in the `<repository>` element in the finding aid. When used in a Bibliographic Reference <bibref>, the name of the party issuing a monograph or other bibliographic work cited in the finding aid.
+   * [`repository`](https://www.loc.gov/ead/tglib/elements/repository.html): The institution or agency responsible for providing intellectual access to the materials being described. Although the repository providing intellectual access usually also has physical custody over the materials, this is not always the case.  When it is clear that the physical custodian does not provide intellectual access, use <physloc> to identify the custodian and <repository> to designate the intellectual caretaker. When a distinction cannot be made, assume that the custodian of the physical objects also provides intellectual access to them and should be recognized as the <repository>.
```
ghukill, line 62: Wonder if mapping EAD `repository` to this new `institution/provider` field might make sense?

jonavellecuerdo: @ghukill @JPrevost 

Would it also be worth to add a mapping from the `publisher` field to the `provider` field for:

1. **Datacite:** The TIMDEX records from JPAL most likely all read "Harvard Dataverse" for this field; records from Zenodo most likely all read "Zenodo"
2. **DspaceDim:** The TIMDEX records from WHOAS most likely all read "Woods Hole Oceanographic Institution" for this field.
3. **SpringshareOaiDc:** For LibGuides, this value will always read "MIT Libraries"
?

jonavellecuerdo:
Hi folks! The **last three commits** are in response @JPrevost 's comments above. In short, updates are made to the following section:

1. **Decision:** Revise the name of the proposed field to `provider` instead, where `provider` refers to "the institution or organization that provides _access_ to the resource described in the TIMDEX record". 
2. **Context:** Exploration of existing fields in non-GIS TIMDEX sources that convey "host institution" in response to @JPrevost 's [comment](https://github.com/MITLibraries/transmogrifier/pull/133#discussion_r1501123480). Main takeaway is that majority of these sources view the "publisher" as an entity that can be the _provider_ of the data as well.
3. **Considerations:** Suggested some additional mappings for non-GIS TIMDEX sources.

Let me know what you think!

docs/adrs/0004-field-for-institution-information.md
```diff
@@ -0,0 +1,125 @@
...
+
+   **Transformer:** OaiDc/Springshare OaiDc
+   **Source(s):** Research Databases (researchdatabases), LibGuides (libguides)
+   * [`<dc:publisher>`](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/publisher): An entity responsible for making the resource available.
+      * **TIMDEX field mapping:**: [`publisher` -> `publication_information`](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/xml/oaidc.py#L91-L94)
+      * **Example value(s):** For Research Databases, `publisher` isn't used; for LibGuides, this value will always read "MIT Libraries".
+
+## Decision
+
+**A new field called "provider" is added to the TIMDEX data model** (i.e., `TimdexRecord.institution`). This field will denote--as its name implies--the institution or organization that provides _access_ to the resource described in the TIMDEX record. This field exists at the top level of the TIMDEX data model, making it easily accessible for referencing in the UI or querying. This solution also avoids using existing fields (discussed below) in ways that obscure the TIMDEX data model.
```
jazairi, line 84: I think `TimdexRecord.institution` should be updated to `TimdexRecord.provider` here.

ghukill, line 84: Should this read "(i.e., `TimdexRecord.provider`)"? Instead of `institution`?

jazairi (changes requested):
See my comment about what I think is an outdated instance of `TimdexRecord.institution`. That small change aside, I'm still 👍 on this. Thanks for adding this context about non-geo sources!

ghukill (approved):
In the spirit of "hopefully this is a widely applicable and longterm solution, but acknowledging future information may require reworking it", I approve of this approach to add a new top level `provider` field.

I think that it will be important to think deeply in Transmogrifier source transformations about the difference between a resource's "publisher" and this proposed "provider", but they do strike me as meaningfully distinct.  And as touched on in comments, err on the side of setting this field only when confidently known or explicitly stated in the record.

In support of this decision, I would also note the field name parity with the [Aardvark field `schema_provider_s`](https://opengeometadata.org/ogm-aardvark/#provider), which is defined as:

> "To clarify which organization **holds the resource** or acts as the custodian for the metadata record and to help users understand which resources they can **access**."

They acknowledge this subtle but important dual purpose of the field where it may denote the institution that holds the actual resource, or just the custodian of the metadata record.  Projecting this "provider" mentality to TIMDEX, I would imagine we utilize either where appropriate for this field, but could lean into the "holds the resource" side of things.  Also noting that their definition of a "provider" very clearly does not imply who _created_ the resource; it's very much access/discovery oriented.

ehanson8 (approved):

jazairi (approved):

PR #136, opened 2024-02-28
Gdt 201 data type refactor

ehanson8:
Implementing the decision described in [ADR #2](https://github.com/MITLibraries/transmogrifier/blob/main/docs/adrs/0002-field-for-data-type-form-information.md)

Run the following command and observe that `Vector data` appears in `content_type`:

```
```

YES

YES

- https://mitlibraries.atlassian.net/browse/GDT-201

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/json/test_aardvark.py
+++ b/tests/sources/json/test_aardvark.py
@@ -46,9 +46,8 @@ def test_aardvark_transform_returns_timdex_record(aardvark_records):
         source_link="https://geodata.libraries.mit.edu/record/abc:123",
         timdex_record_id="cool-repo:123",
         title="Test title 1",
-        citation="Test title 1. Geospatial data. "
-        "https://geodata.libraries.mit.edu/record/abc:123",
-        content_type=["Geospatial data"],
+        citation="Test title 1. https://geodata.libraries.mit.edu/record/abc:123",
+        content_type=["Not specified"],
         rights=[timdex.Rights(description="Access rights", kind="Access rights")],
         links=[
             timdex.Link(
@@ -66,6 +65,7 @@ def test_aardvark_get_optional_fields_non_field_method_values_success(
 ):
     transformer = MITAardvark("cool-repo", aardvark_record_all_fields)
     record = next(transformer)
+    assert record.content_type == ["Vector data"]
     assert record.format == "Shapefile"
     assert record.languages == ["eng"]
     assert record.summary == ["A description"]
@@ -378,7 +378,6 @@ def test_aardvark_get_subjects_success(aardvark_record_all_fields):
         timdex.Subject(value=["Geography"], kind="Dublin Core; Subject"),
         timdex.Subject(value=["Earth"], kind="Dublin Core; Subject"),
         timdex.Subject(value=["Dataset"], kind="Subject scheme not provided"),
-        timdex.Subject(value=["Vector data"], kind="Subject scheme not provided"),
     ]
 
 
--- a/transmogrifier/sources/json/aardvark.py
+++ b/transmogrifier/sources/json/aardvark.py
@@ -126,7 +126,7 @@ class MITAardvark(JSONTransformer):
         fields["alternate_titles"] = self.get_alternate_titles(source_record) or None
 
         # content_type
-        fields["content_type"] = ["Geospatial data"]
+        fields["content_type"] = source_record.get("gbl_resourceType_sm")
 
         # contributors
         fields["contributors"] = self.get_contributors(source_record) or None
@@ -456,7 +456,6 @@ class MITAardvark(JSONTransformer):
             "dct_spatial_sm": "Dublin Core; Spatial",
             "dct_subject_sm": "Dublin Core; Subject",
             "gbl_resourceClass_sm": "Subject scheme not provided",
-            "gbl_resourceType_sm": "Subject scheme not provided",
         }
 
         for subject_field, subject_kind in aardvark_subject_fields.items():
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -36,18 +36,18 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:4052031f9ac18924e94be7c30a5f0af5843a4752b8c8cb9034cd978be252b61b",
-                "sha256:53897701ab4f307fbcfdade673eae809dfc5eabb6102053c84907aa27de66e53"
+                "sha256:66303b5f26d92afb72656ff490b22ea72dfff8bf1a29e4a0c5d5f11ec56245dd",
+                "sha256:898ad2123b18cae8efd85adc56ac2d1925be54592aebc237020d4f16e9a9e7a9"
             ],
-            "version": "==1.34.35"
+            "version": "==1.34.52"
         },
         "botocore": {
             "hashes": [
-                "sha256:8a2b53ab772584a5f7e2fe1e4a59028b0602cfef8e39d622db7c6b670e4b1ee6",
-                "sha256:b67b8c865973202dc655a493317ae14b33d115e49ed6960874eb05d950167b37"
+                "sha256:05567d8aba344826060481ea309555432c96f0febe22bee7cf5a3b6d3a03cec8",
+                "sha256:187da93aec3f2e87d8a31eced16fa2cb9c71fe2d69b0a797f9f7a9220f5bf7ae"
             ],
             "markers": "python_version >= '3.8'",
-            "version": "==1.34.35"
+            "version": "==1.34.52"
         },
         "certifi": {
             "hashes": [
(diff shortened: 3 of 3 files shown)


tests/sources/json/test_aardvark.py
```diff
@@ -46,9 +46,8 @@ def test_aardvark_transform_returns_timdex_record(aardvark_records):
         source_link="https://geodata.libraries.mit.edu/record/abc:123",
         timdex_record_id="cool-repo:123",
         title="Test title 1",
-        citation="Test title 1. Geospatial data. "
-        "https://geodata.libraries.mit.edu/record/abc:123",
-        content_type=["Geospatial data"],
+        citation="Test title 1. https://geodata.libraries.mit.edu/record/abc:123",
+        content_type=["Not specified"],
```
ghukill, line 50: I'm curious if a value of "Not specified" is what we want for the `content_type`?  

Took me a minute to track it down, but looks like this is [applied by the base `Transformer` class here](https://github.com/MITLibraries/transmogrifier/blob/GDT-201-data-type-refactor/transmogrifier/sources/transformer.py#L258-L259) if the source transformer does not set it.

When `content_type` was considered a higher level type (if maybe somewhat variable across sources), I think this made sense.  But if we interpret it as a slightly lower level type, where it might be okay to be `NULL`, we might want to consider removing this logic from the base Transformer and allowing it to be `NULL`.

Taking `content_type` [examples from the table in the ADR](https://github.com/MITLibraries/transmogrifier/blob/main/docs/adrs/0002-field-for-data-type-form-information.md#option-5---map-format-to-format-filter-map-content_type-to-data-type-filter), I'm not sure what we gain by having `"Not specified"` showing up in those values.  

I _do_ think this might make sense for the proposed, higher level `content_class`, and that probably should be a required field as well.  But as it stands, I believe that `content_type` is _not_ required, and therefore we could just leave it set as `None` or an empty list `[]`.

Thoughts?

tests/sources/json/test_aardvark.py
```diff
@@ -66,6 +65,7 @@ def test_aardvark_get_optional_fields_non_field_method_values_success(
 ):
     transformer = MITAardvark("cool-repo", aardvark_record_all_fields)
     record = next(transformer)
+    assert record.content_type == ["Vector data"]
```
ghukill, line 68: 🥳

transmogrifier/sources/json/aardvark.py
```diff
@@ -126,7 +126,7 @@ def get_optional_fields(self, source_record: dict) -> dict | None:
         fields["alternate_titles"] = self.get_alternate_titles(source_record) or None
 
         # content_type
-        fields["content_type"] = ["Geospatial data"]
+        fields["content_type"] = source_record.get("gbl_resourceType_sm")
```
ghukill, line 129: Ignoring the comment above about whether this be _allowed_ remain `None` or a potentially empty list, I think it's a good sign that this change was so slight.  Feels like the transformer for Aardvark records is flexible in a good way.

ghukill (commented):
I think the mapping change looks good.  But I left a comment for discussion about whether or not we want to keep `["Not specified"]` as a default value for records now that we're slightly reimagining `content_type`.

I would observe that this would have bearing on other sources: if we do remove this default behavior, then other sources would potentially "lose" this default value for `content_type`, which to me would be okay.

That said, barring a decision there, looking good.

tests/sources/json/test_aardvark.py
ghukill: If helpful for discussion, here is a breakdown of the `format` / `content_type` combinations from MIT data:
```
┌───────────────────────────────────────┬───────────┬─────────────┐
│             content_type              │  format   │ combo_count │
│               varchar[]               │  varchar  │    int64    │
├───────────────────────────────────────┼───────────┼─────────────┤
│ [Polygon data]                        │ Shapefile │        1089 │
│ [Point data]                          │ Shapefile │         620 │
│ [Line data]                           │ Shapefile │         293 │
│ [Polygon data, Point data]            │ Shapefile │          33 │
│ [Mixed]                               │           │          10 │
│ [Line data, Point data]               │ Shapefile │           4 │
│ [Not specified]                       │ GeoTIFF   │           3 │
│ [Not specified]                       │           │           3 │
│ [Line data, Point data, Polygon data] │ Shapefile │           1 │
│ [Polygon data, Point data, Mixed]     │ Shapefile │           1 │
│ [Mixed]                               │ Shapefile │           1 │
├───────────────────────────────────────┴───────────┴─────────────┤
│ 11 rows                                               3 columns │
└─────────────────────────────────────────────────────────────────┘
```

And OGM data (with some of the long tail omitted in the results, but looking at all 120k records):
```
┌───────────────────────────────────────────────┬────────────┬─────────────┐
│                 content_type                  │   format   │ combo_count │
│                   varchar[]                   │  varchar   │    int64    │
├───────────────────────────────────────────────┼────────────┼─────────────┤
│ [Image data]                                  │ TIFF       │       30775 │
│ [Polygon data]                                │ Shapefile  │       26702 │
│ [Raster data]                                 │ GeoTIFF    │       10084 │
│ [Line data]                                   │ Shapefile  │        8146 │
│ [Not specified]                               │ JPEG       │        7094 │
│ [Point data]                                  │ Shapefile  │        7090 │
│ [Topographic maps, Military maps]             │ GeoTIFF    │        4129 │
│ [Not specified]                               │ GeoTIFF    │        4004 │
│ [Fire insurance maps]                         │ JPEG       │        3871 │
│ [Not specified]                               │ PDF        │        3009 │
│ [Image data]                                  │            │        2185 │
│ [Cadastral maps]                              │ JPEG       │        1782 │
│ [Atlases]                                     │ JPEG       │        1245 │
│ [Raster data]                                 │ MrSID      │        1233 │
│ [Not specified]                               │ Mixed      │        1223 │
│ [Geological maps]                             │ JPEG       │        1216 │
│ [Raster data]                                 │ TIFF       │         916 │
│ [Aerial photographs]                          │ JPEG       │         680 │
│ [Not specified]                               │            │         517 │
│ [Thematic maps]                               │ JPEG       │         453 │
│       ·                                       │  ·         │           · │
│       ·                                       │  ·         │           · │
│       ·                                       │  ·         │           · │
│ [Raster data]                                 │ PDF        │           1 │
│ [Early maps, World maps]                      │ JPEG2000   │           1 │
│ [Zoning maps, Thematic maps]                  │ TIFF       │           1 │
│ [Mine maps, Topographic maps]                 │ JPEG       │           1 │
│ [Pictorial maps, Tourist maps]                │ JPEG       │           1 │
│ [Aerial views, Pictorial maps, Thematic maps] │ JPEG       │           1 │
│ [Not specified]                               │ GeoJSON    │           1 │
│ [Table data, Line data]                       │ Shapefile  │           1 │
│ [Vector data, Raster data]                    │ Shapefile  │           1 │
│ [Vector data, Point data, Polygon data]       │ Shapefile  │           1 │
│ [Nautical charts]                             │ JPEG2000   │           1 │
│ [Early maps, Atlases]                         │ JPEG       │           1 │
│ [Vector data, Point data, Polygon data]       │ GeoPackage │           1 │
│ [Table data, Point data]                      │            │           1 │
│ [Cadastral maps, Topographic maps]            │ JPEG       │           1 │
│ [Multi-spectral data]                         │ Mixed      │           1 │
│ [Bathymetric maps, Thematic maps]             │ JPEG       │           1 │
│ [Thematic maps, Fire insurance maps]          │ JPEG       │           1 │
│ [Topographic maps]                            │            │           1 │
│ [Multi-spectral data]                         │ Shapefile  │           1 │
├───────────────────────────────────────────────┴────────────┴─────────────┤
│ 168 rows (40 shown)                                            3 columns │
└──────────────────────────────────────────────────────────────────────────┘
```

ehanson8: I agree that `Not specified` no longer serves the same purpose but I would be inclined to group that change with the creation of the `content_class` field rather than make the change here. That feels like a significant data model change (which I would feel about almost changes to the `Transformer`class) to throw into this PR without having a fully considered what the replacement looks like and the impact on all the sources. So, yes but after we're ready to properly replace it!

ghukill: I might pushback a bit that this is a significant data model change, insofar as it's just removing a default value of "Not specified" when the transformer class does not explicitly set it.  By removing this logic, we don't modify any values, just lose this default.

Ran a check against all sources in Dev1, looks like 17,273 records have "Not specified", and here's a breakdown by source:
```
+----------+------------+
|source    |count_star()|
+----------+------------+
|MIT Alma  |7913        |
|DSpace@MIT|9360        |
+----------+------------+
```

But, to your point, it does feel like that change would be more aligned with the `content_class` (name still TBD) work.  Happy to wait and roll into a different PR!  Thanks for considering here.

ghukill (approved):
Given some discussion about removing "Not specified" as part of a different PR, approving this one!

jonavellecuerdo (approved):
Looks good! As discussed, there will be a separate effort for the addition of a `content_class` field.

**Also**, PR will need to be rebased with `main` prior to merging, I think.

PR #141, opened 2024-03-08
Gdt 210 publishers refactor

ehanson8:
Implement changes specified in [ADR # 3](https://github.com/MITLibraries/transmogrifier/blob/main/docs/adrs/0003-support-aggregations-on-publisher-name.md) 

Run the following command and observe that `dates` and `locations` contain values from the `publishers` field

```
pipenv run transform -i tests/fixtures/marc/marc_record_all_fields.xml -o output/alma-transformed-records.json -s alma
```

YES

YES

- https://mitlibraries.atlassian.net/browse/GDT-210

### Developer

### Code Reviewer(s)


transmogrifier/sources/xml/datacite.py
```diff
@@ -255,6 +255,15 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
                 source_record_id,
             )
 
+        # publishers
```
ehanson8, line 249: I considered but decided not to start breaking this out as a field method in the older transforms, I think that work will benefit from a comprehensive rather than piecemeal approach

transmogrifier/sources/transformer.py
```diff
@@ -364,6 +369,30 @@ def get_optional_fields(self, _source_record: dict[str, JSON] | Tag) -> dict | N
         """
         return {}
 
+    @final
+    @staticmethod
+    def create_dates_and_locations_from_publishers(fields: dict) -> dict:
```
ehanson8, line 372: This is specific to this use case but as we do more of these duplicate objects for facets, this can be refactored to be more generic

ghukill: I think this is an interesting case.

I'm on board with waiting to either make more generic, establish some kind of `post_transform_hooks` linkage, and/or refactor to use metadata field methods.

An example of how these things might be intertwined.  We could create decorators that are _optionally_ applied to field methods, that would do that work

```python
# post field method run decorators
def extract_dates():
    pass

def extract_locations():
    pass

def extract_subjects():
    pass

class MyTransformer():
    
    #specific field method getting those decorators
    @extract_dates
    @extract_locations
    def get_publishers(self):
        pass
```

transmogrifier/sources/transformer.py
```diff
@@ -251,17 +253,20 @@ def _transform(self, source_record: dict[str, JSON] | Tag) -> TimdexRecord | Non
             **self.get_required_fields(source_record),
             **optional_fields,
         }
+        fields = self.create_dates_and_locations_from_publishers(fields)
```
ghukill, line 254: It's a minor nit, but I'd propose a blank line above this method call and after the `fields = {...}` call.  That previous call is a MAJOR one, where the bulk of the transform happens.  

This `self.create_dates_and_locations_from_publishers()` call -- as discussed below -- is probably pretty temporary, and secondary, and feels like it needs some space.  

I was going to propose some commenting in this method, but the more I look at it, and the more we kind of push its current architecture, it seems clear that a refactor will rework this fairly heavily at some point.

ghukill (changes requested):
Looking real good!  Easy to follow the updates, and the logic looks clean throughout.

I liked the approach of a new commit for adding the field `publishers` and a separate commit to remove the field `publication_information`.  Good idea.

Left a couple of comments/requests, but mostly just comments.  The request -- and please pushback if missing something -- was considering `Location@kind="Place of Publication"` instead of `Locations@kind="Published"`

But overall, looks real good to me; deft touch at some fairly invasive changes.

transmogrifier/helpers.py
```diff
@@ -28,13 +27,10 @@ def generate_citation(extracted_data: dict) -> str:
...
-        publication_dates = None
-    else:
-        publication_dates = [
-            date.value
-            for date in extracted_data["dates"]
-            if date.kind == "Publication date" and date.value
+    publication_dates = ""
+    if dates := extracted_data.get("dates"):
+        publication_dates = [  # type: ignore[assignment]
+            date.value for date in dates if date.kind == "Publication date" and date.value
```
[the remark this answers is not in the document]
ehanson8: I thought we had this but it was just these 2 more specific tickets: [TIMX-117](https://mitlibraries.atlassian.net/browse/TIMX-117) and [TIMX-138](https://mitlibraries.atlassian.net/browse/TIMX-138). I'll create a stub ticket to re-evaluate it completely. It's too tied to RDI needs that don't suit TIMDEX's broader purpose now

transmogrifier/sources/transformer.py
```diff
@@ -19,9 +19,9 @@
 # should not be a security issue.
 from lxml import etree  # nosec B410
 
+import transmogrifier.models as timdex
 from transmogrifier.config import SOURCES
 from transmogrifier.helpers import DeletedRecordEvent, generate_citation
-from transmogrifier.models import TimdexRecord
```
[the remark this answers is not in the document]
ehanson8: I was looking to standardize with the import structure in `transformer` with all the other sources transforms which use the `import transmogrifier.models as timdex` pattern. But I can change this to explicit imports if you think that's a better fit

transmogrifier/sources/transformer.py
ehanson8: Good call, and agree on the need for a future refactor

transmogrifier/sources/transformer.py
```diff
@@ -364,6 +369,30 @@ def get_optional_fields(self, _source_record: dict[str, JSON] | Tag) -> dict | N
...
+    @final
+    @staticmethod
+    def create_dates_and_locations_from_publishers(fields: dict) -> dict:
+        """Add Date and Location objects based on data in publishers field.
+
+        Args:
+            fields: A dict of fields representing a TIMDEX record.
+        """
+        if fields.get("publishers"):
+            for publisher in fields["publishers"]:
```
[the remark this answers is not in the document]
ehanson8: Great suggestion!

transmogrifier/sources/transformer.py
ehanson8: I think the decorator approach has promise whenever we tackle this

transmogrifier/sources/transformer.py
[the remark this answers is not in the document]
ghukill: @ehanson8 - ah, I never really connected those dots!   I never stopped to consider the `timdex.<field>` pattern.  I understood it came from `TIMDEXRecord`, but hadn't really interrogated it.

Definitely not worth changing it now, but even though more text, I might propospe that we make those:

```python
from transmogrifier.models import TIMDEXRecord

# then in code...
fields["foo"] = TIMDEXRecord.foo(...)
```

It's just real explicit that way.

I'm fine either way.  If you want to keep this change such that mimicks other sources, makes sense to me.  Or, if we want to keep this as-is, lean into explicit imports, with the idea we might move others that direction, also works.

My main issue with the `import timdex.models as timdex` is that it's kind of opaque, hides that `timdex` is in context.

transmogrifier/sources/xml/datacite.py
```diff
@@ -255,6 +255,15 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+        if publisher := xml.metadata.find("publisher", string=True):
```
jonavellecuerdo, line 250: [Non-blocking] Initially, I was wondering this would work for records where there are multiple XML elements tagged with "publisher", but I remembered that `Datacite` is a transformer used for the following sources:
* `zenodo`: Every record has a single `publisher` element ("Zenodo")
* `jpal`: Every record has a single `publisher` element ("Harvard Dataverse")
This was explored as part of[ this comment on ADR-0003](https://github.com/MITLibraries/transmogrifier/pull/132#discussion_r1507657483).

Just noting that this might be useful information to keep somewhere (maybe we can start a new Markdown file in `docs/` at some point! This Markdown file could share a link to the Datacite schema or just a description of this finding.

transmogrifier/sources/xml/dspace_mets.py
```diff
@@ -151,6 +151,12 @@ def get_optional_fields(self, xml: Tag) -> dict:
         for publisher in xml.find_all("mods:publisher", string=True):
             fields.setdefault("publication_information", []).append(publisher.string)
 
+        # publishers
```
jonavellecuerdo, line 150: Similar comment to above. How do we know if a source has multiple elements tagged with `mods:publisher`? 🤔

transmogrifier/sources/transformer.py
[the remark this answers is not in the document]
ehanson8: To clarify, `timdex` is used for calling objects defined in `models.py` (`Contributor`, ` Date`, `Location`, etc.)  not in `TIMDEXRecord`. Those are used by the `TIMDEXRecord` fields but I think we'd need to establish the field method model we've discussed  before using `fields["foo"] = TIMDEXRecord.foo(..)` (which you are probably implying here). I'll leave my changes in place for now, I think it's something to address globally when we're done with GDT, I just want consistency across the repo, whichever way we do it.

transmogrifier/sources/xml/datacite.py
ehanson8: I think we can have a big discussion about documentation for all the sources but I think that's beyond the scope of this PR

transmogrifier/sources/xml/dspace_mets.py
ehanson8: We wouldn't know but this transform is set up handle multiple publisher elements

jonavellecuerdo:
@ehanson8 Thank you for working on this and organizing the commits to make for an easier review! I agree with all of @ghukill 's suggestions above and acknowledge that some of the proposed updates will be implemented at a later iteration. This looks good to me, but I'll defer to @ghukill for final approval!

transmogrifier/sources/transformer.py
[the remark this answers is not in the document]
ghukill: @ehanson8 - makes sense to me, and good clarification about `timdex.*` being used for all the field classes.  Keeping this change you made feels like it brings it inline with other sources, and then could evaluate globally at some point.

PR #146, opened 2024-03-11
Map Aardvark.dct_spatial_sm to Locations@kind="Place Name"

jonavellecuerdo:
Map `Aardvark.dct_spatial_sm` to `Locations@kind="Place Name"` to support the use of place names
for location-based searching for GeoData. 

**Note:** `MITAardvark` is the only source transformer that writes `Subject@kind="Dublin Core; Spatial"` (i.e., what `Aardvark.dct_spatial_sm` currently maps to), so the updated mapping is only applied to this transformer.

1. Temporarily set AWS credentials for `TimdexManagers` for `Dev1` in your terminal.
2. Transform `gismit` records
   * Run the following command from your local clone of `transmogrifier`:
      ```
      pipenv run transform -i s3://timdex-extract-dev-222053980223/gismit/gismit-2024-03-11-full-extracted-records-to-index.jsonl -o output/output_gismit.json -s gismit -v
      ```
   * View `output/output_gismit.json`. Lines 67-88 should read:
      image

  
3. Transform `gisogm` records (a partial run, at least!)
   * Run the following command from your local clone of `transmogrifier`:
      ```
      pipenv run transform -i s3://timdex-extract-dev-222053980223/gisogm/gisogm-2024-03-01-full-extracted-records-to-index.jsonl -o output/output_gisogm.json -s gisogm -v
      ```
   * View `output/output_gisogm.json`. Lines 49-66 should read:
     image

NO

NO

https://mitlibraries.atlassian.net/browse/GDT-217

### Developer

### Code Reviewer(s)

### DIFF
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -402,22 +402,16 @@ class Transformer(ABC):
         Args:
            fields: A dict of fields representing a TIMDEX record.
         """
-        if not (subjects := fields.get("subjects")):
+        if fields.get("subjects") is None:
             return fields
 
-        if not (
-            spatial_subjects := list(
-                filter(
-                    lambda subject: subject.kind == "Dublin Core; Spatial", subjects  # type: ignore[arg-type]
-                )
-            )
-        ):
-            return fields
+        spatial_subjects = [
+            subject
+            for subject in fields.get("subjects", [])  # defaults to empty list
+            if subject.kind == "Dublin Core; Spatial" and subject.value is not None
+        ]
 
         for subject in spatial_subjects:
-            if not subject.value:
-                continue
-
             for place_name in subject.value:
                 subject_location = timdex.Location(value=place_name, kind="Place Name")
                 if subject_location not in fields.setdefault("locations", []):
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -253,6 +253,7 @@ class Transformer(ABC):
         }
 
         fields = self.create_dates_and_locations_from_publishers(fields)
+        fields = self.create_locations_from_spatial_subjects(fields)
 
         # If citation field was not present, generate citation from other fields
         if fields.get("citation") is None:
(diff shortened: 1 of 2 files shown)


ghukill (changes requested):
The update looks good, and I think achieves what we need, but I might propose that we lean into the post-transform "hook" or "enrichment" (name TBD) pattern that was established in https://github.com/MITLibraries/transmogrifier/pull/141.

While I think it's agreed that we'll want to revist how to orchestrate these hooks, this second case of needing to do something very similar suggests there is a pattern that we can identify and abstract out.

Until then, I think we'd benefit from not locating this `Subject --> Location` data extraction/duplication in the `MITAardvark` transform only, lest we forget about it later and accidentally duplicate it's functionality.  Moreover, if we lean into this pattern now and define it at the `Transformer` level, I think it provides more concrete use cases to consider when that refactor happens in earnest.

More concretely, where we currently have:
```python
fields = self.create_dates_and_locations_from_publishers(fields)
```

I might propose we add another like:
```python
# post transform hooks and enrichments
fields = self.create_dates_and_locations_from_publishers(fields)
fields = self.create_locations_from_spatial_subjects(fields)  # <--- new
```

The needle I think we want to thread is: making changes now that are solid and understandable, while understanding a future refactor may combine and extend this pattern that is emerging from these first two use cases.

jonavellecuerdo:
@ghukill Thank you for clarifying! I undid the changes to the `MITAardvark` transformer and implemented the solution as a post-transform hook in the `transformer` module.

transmogrifier/sources/transformer.py
```diff
@@ -392,3 +393,33 @@ def create_dates_and_locations_from_publishers(fields: dict) -> dict:
...
+            return fields
+
+        if not (
+            spatial_subjects := list(
+                filter(
+                    lambda subject: subject.kind == "Dublin Core; Spatial", subjects  # type: ignore[arg-type]
+                )
+            )
+        ):
+            return fields
```
ghukill, line 415: I think that I see the logic here, refining a list of potential subjects down using `filter` + a lambda, but perhaps this could be a simpler list comprehension?

```python
# early return; handles edge case where subjects is value None and would return as such from .get()
if not fields.get("subjects"):
    return fields

# filters list, potentially to empty list which is okay
spatial_subjects = [
    subject
    for subject in fields.get("subjects", [])
    if subject.kind == "Dublin Core; Spatial" and subject.value is not None
]
```

ghukill (commented):
General structure, looks good!  Thanks for going with this pattern of locating these "post hook" work together.

I left a syntactical / stylistic comment for the filtering of `subjects` down to spatial ones.    I find list comprehensions easier to scan than `filter` + lambda, but open to pushback!

transmogrifier/sources/transformer.py
ehanson8: Agree on list comprehension, that's what we've used throughout the app and I find it more readable as well

ghukill (approved):
Looks good!

ehanson8 (approved):
Looks good!

PR #173, opened 2024-05-09
[wip] rewrite DSpaceDim.get_dates using lxml

jonavellecuerdo:
Describe the overall purpose of the PR changes and any useful background context.

Explain how to see the proposed changes in the application if possible.

Delete this section if it isn't applicable to the PR.

YES | NO

YES | NO

- Include links to Jira Software and/or Jira Service Management tickets here.

### Developer

### Code Reviewer(s)


transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -224,59 +229,58 @@ def get_optional_fields(self, source_record: Tag) -> dict | None:
...
-                qualifier="tableofcontents",
-                string=True,
-            )
-        ]
+    def get_contents(cls, source_record: etree._Element) -> list[str]:
+        contents = cls.string_list_from_xpath(
+            source_record,
+            ".//dim:field[@element='description' and @qualifier='tableofcontents']",
+        )
+        return contents
```
ghukill, line 246: This to me is a really interesting one.

@jonavellecuerdo, I think in a slack huddle I incorrectly thought that `@element` was Xpath for matching the actual element; now I realize it's an actual XML attribute called `element`!

With that in mind -- personally -- I find that BS4 kind of obfuscates this a bit.  Yes it may seem handy there are `element=...` and `qualifier=...` kwargs on the `find_all()` method, but at a glance it's not obvious those are part of the XML structure itself.  

I find the Xpath is more self-describing and explicit in this way.

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -224,59 +229,58 @@ def get_optional_fields(self, source_record: Tag) -> dict | None:
...
 
     @classmethod
     def _parse_date_elements(
-        cls, source_record: Tag, source_record_id: str
+        cls, source_record: etree._Element, source_record_id: str
     ) -> Iterator[timdex.Date]:
-        for date_element in source_record.find_all(
-            "dim:field", element="date", string=True
+        for date_element in cls.xpath_query(
+            source_record, ".//dim:field[@element='date']"
```
ghukill, line 262: Noting here, but think it's true throughout: probably don't need `.//` prefixing all Xpath queries.  If the `.xpath()` is called from the root element, then `//` would be sufficient to find an element anywhere in the XML file, as all elements are children of this.

Fairly minor, but felt worth mentioning.

ghukill:
Kind of a general comment. 

I think `dspace_dim` is a _bit_ unusual, given that it's kind of a flat structure with `@element` attributes on repeating `dim:field` elements.  On one hand, I don't think this exercises XPath's ability to dig deep into structures.  On the other hand, it's medium-common for our records, so it's a good to compare against BS4 which is decently well suited to handle this.

jonavellecuerdo:
TBD

PR #175, opened 2024-05-16
Update exception handling

ehanson8:
Adding more explicit flow control through exception handling in the `Transformer` class. Previously, records were skipped when `get_optional_fields` returned `None`. Adding a `SkippedRecordEvent` exception that can handle records that should be skipped for an invalid `content_type` or records with significant structural issues that raise the risk of unknown exceptions later in the process. This is a minor update to the orchestration that should aid us with the field method refactor.

As we proceed with the refactor in other transforms, use the `SkippedRecordEvent` exception whenever `get_optional_fields` returns `None`. Logging can be added as necessary depending on what triggered the skip.

The `Datacite` class was updated to raise  the`SkippedRecordEvent` exception illustrating that tests such as `test_zenodo_skips_records_with_invalid_content_types` still pass.

NO

NO

- NA

### Developer

### Code Reviewer(s)

### DIFF
--- a/transmogrifier/exceptions.py
+++ b/transmogrifier/exceptions.py
@@ -16,6 +16,6 @@ class SkippedRecordEvent(Exception):  # noqa: N818
         source_record_id: The ID for the source record.
     """
 
-    def __init__(self, source_record_id: str | None, message: str) -> None:
+    def __init__(self, message: str | None = None, source_record_id: str | None = None):
+        super().__init__(message)
         self.source_record_id = source_record_id
-        self.message = message
--- a/transmogrifier/exceptions.py
+++ b/transmogrifier/exceptions.py
@@ -1,0 +1,20 @@
+class DeletedRecordEvent(Exception):  # noqa: N818
+    """Exception raised for records with a deleted status.
+
+    Attributes:
+        timdex_record_id: The TIMDEX record ID (not the source record ID) for the record.
+    """
+
+    def __init__(self, timdex_record_id: str) -> None:
+        self.timdex_record_id = timdex_record_id
+
+
+class SkippedRecordEvent(Exception):  # noqa: N818
+    """Exception raised for records that should be skipped.
+
+    Attributes:
+        source_record_id: The ID for the source record.
+    """
+
+    def __init__(self, source_record_id: str) -> None:
+        self.source_record_id = source_record_id
--- a/transmogrifier/exceptions.py
+++ b/transmogrifier/exceptions.py
@@ -16,5 +16,6 @@ class SkippedRecordEvent(Exception):  # noqa: N818
         source_record_id: The ID for the source record.
     """
 
-    def __init__(self, source_record_id: str) -> None:
+    def __init__(self, source_record_id: str | None, message: str) -> None:
         self.source_record_id = source_record_id
+        self.message = message
--- a/transmogrifier/helpers.py
+++ b/transmogrifier/helpers.py
@@ -132,15 +132,3 @@ def validate_date_range(
         end_date,
     )
     return False
-
-
-class DeletedRecordEvent(Exception):  # noqa: N818
-    """Exception raised for records with a deleted status.
-
-    Attributes:
-        timdex_record_id: The TIMDEX record ID (not the source record ID) for the record
-
-    """
-
-    def __init__(self, timdex_record_id: str) -> None:
-        self.timdex_record_id = timdex_record_id
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -72,11 +72,11 @@ class Transformer(ABC):
             except SkippedRecordEvent:
                 self.skipped_record_count += 1
                 continue
-            if record:
-                self.transformed_record_count += 1
-                return record
-            self.skipped_record_count += 1
-            continue
+            if not record:
+                self.skipped_record_count += 1
+                continue
+            self.transformed_record_count += 1
+            return record
 
     @final
     def transform_and_write_output_files(self, output_file: str) -> None:
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -17,7 +17,8 @@ from attrs import asdict
 # should not be a security issue.
 import transmogrifier.models as timdex
 from transmogrifier.config import SOURCES
-from transmogrifier.helpers import DeletedRecordEvent, generate_citation, validate_date
+from transmogrifier.exceptions import DeletedRecordEvent, SkippedRecordEvent
+from transmogrifier.helpers import generate_citation, validate_date
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
@@ -67,6 +68,9 @@ class Transformer(ABC):
                 record = self.transform(source_record)
             except DeletedRecordEvent as error:
                 self.deleted_records.append(error.timdex_record_id)
+                continue
+            except SkippedRecordEvent:
+                self.skipped_record_count += 1
                 continue
             if record:
                 self.transformed_record_count += 1
--- a/transmogrifier/sources/xml/datacite.py
+++ b/transmogrifier/sources/xml/datacite.py
@@ -56,7 +56,7 @@ class Datacite(XMLTransformer):
                     fields["content_type"] = [content_type]
                 else:
                     message = f'Record skipped based on content type: "{content_type}"'
-                    raise SkippedRecordEvent(source_record_id, message)
+                    raise SkippedRecordEvent(message, source_record_id)
         else:
             logger.warning(
                 "Datacite record %s missing required Datacite field resourceType",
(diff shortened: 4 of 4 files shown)


transmogrifier/helpers.py
```diff
@@ -132,15 +132,3 @@ def validate_date_range(
...
-class DeletedRecordEvent(Exception):  # noqa: N818
-    """Exception raised for records with a deleted status.
-
-    Attributes:
-        timdex_record_id: The TIMDEX record ID (not the source record ID) for the record
-
-    """
-
-    def __init__(self, timdex_record_id: str) -> None:
-        self.timdex_record_id = timdex_record_id
```
ghukill, line 146: +1 to the move

transmogrifier/sources/transformer.py
```diff
@@ -68,6 +69,9 @@ def __next__(self) -> timdex.TimdexRecord:
             except DeletedRecordEvent as error:
                 self.deleted_records.append(error.timdex_record_id)
                 continue
+            except SkippedRecordEvent:
+                self.skipped_record_count += 1
+                continue
```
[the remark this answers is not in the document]
ghukill: I'm unable to comment on the lines below, as they did not change, but thoughts on updating the logic to treat a successful record as the default path?  

I know this will get touched on in orchestration updates, but while touching this `__next__` method, thought it could be a good time to nudge it.

e.g.
```python
if not record:
    continue
return record  
```

transmogrifier/sources/xml/datacite.py
```diff
@@ -54,7 +55,7 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
                 if self.valid_content_types([content_type]):
                     fields["content_type"] = [content_type]
                 else:
-                    return None
+                    raise SkippedRecordEvent(source_record_id)
```
ghukill, line 58: Looks good.  But see comment above about adding messaging close to the error/situation, confident it would get logged higher up.

transmogrifier/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SkippedRecordEvent(Exception):  # noqa: N818
+    """Exception raised for records that should be skipped.
+
+    Attributes:
+        source_record_id: The ID for the source record.
+    """
+
+    def __init__(self, source_record_id: str) -> None:
+        self.source_record_id = source_record_id
```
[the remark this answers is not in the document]
ehanson8: Good point, I'll do that

transmogrifier/sources/transformer.py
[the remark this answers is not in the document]
ehanson8: Good call on the exception message and I was frustrated with the repetition in that method so thank you very much for such a simple refactor!

ehanson8:
@ghukill Made changes in response to your review. Also, didn't realize it's possible to change the target branch on an existing PR so I'll keep this one open

transmogrifier/exceptions.py
```diff
@@ -0,0 +1,21 @@
...
+class SkippedRecordEvent(Exception):  # noqa: N818
+    """Exception raised for records that should be skipped.
+
+    Attributes:
+        source_record_id: The ID for the source record.
+    """
+
+    def __init__(self, source_record_id: str | None, message: str) -> None:
+        self.source_record_id = source_record_id
+        self.message = message
```
ghukill, line 21: I think there is some nuance to consider here, but feels like the time to think about these exceptions as a) they are now going into `main`, and b) might be fairly central to new orchestration patterns when implemented.

I would propose two things:
1. We can instantiate these exceptions just with a string message if we like, just like normal exceptions, and have this represented when you stringify the exception object
2. All _extra_ properties are optional, unless we decide they should be required

Example of how this could be done:
```python
class SkippedRecordEvent(Exception):
    """"""
    def __init__(self, message:str|None=None, source_record_id:str|None=None):
        super().__init__(message) # calls base Exception constructor
        self.source_record_id = source_record_id
```

By default, the base `Exception` class is expecting a positional argument `message` that is then used in the `__str__` method to stringify it.

This allows some behavior like this:
```
# acts like other exceptions, allows string message as first and only argument
In [19]: e = SkippedRecordEvent("Hello world!")

In [20]: str(e)
Out[20]: 'Hello world!'

In [21]: e
Out[21]: __main__.SkippedRecordEvent('Hello world!')

In [22]: e.source_record_id
# None passed

# also supports extra properties if defined
In [23]: e2 = SkippedRecordEvent("Hello world!", source_record_id="acb123")

In [24]: str(e2)
Out[24]: 'Hello world!'

# source_record_id also present if we want to use it
In [25]: e2.source_record_id
Out[25]: 'acb123'

# not ideal, but like default exceptions, can even raise it without messages
In [26]: e3 = SkippedRecordEvent()

In [27]: str(e3)
Out[27]: 'None' # string of None

In [28]: e3.source_record_id
# also None
```

Two additional considerations, maybe now, maybe later:
1. Does this have implications for `DeletedRecordEvent`?  Do we want to update patterns so it has similar default, predictable behavior?
2. Do we want more granular exception classes like `InvalidRecordError` that extend `SkippedRecordEvent`, such that we could raise those more meaningful, granular exceptions from code, but still have a single `except SkippedRecordEvent: ...` logic at the higher level to group skipping behavior for all those child Exception types?

Both seem like they could wait.  And for that matter, the suggestions above could too.  But felt like a good time to raise them.

transmogrifier/sources/xml/datacite.py
```diff
@@ -54,7 +55,8 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
                 if self.valid_content_types([content_type]):
                     fields["content_type"] = [content_type]
                 else:
-                    return None
+                    message = f'Record skipped based on content type: "{content_type}"'
+                    raise SkippedRecordEvent(source_record_id, message)
```
ghukill, line 59: See comments above about possible reworking of these custom exceptions.

Even if we don't go that route, I think perhaps the `message` should be the first argument, and positional, and then `source_record_id` be a named argument is optional.

ghukill (commented):
Looking good, but left some comments on the exceptions.

As noted, it feels like a time to iron out what they'll look and feel like, if they do become more central to things.  

I could be persuaded that no changes are needed, but curious your thoughts!

transmogrifier/exceptions.py
ehanson8: Agree on the proposed exception formatting!

Re: additional considerations, those feel better to address during the full orchestration discussion. It doesn't feel like we have full use cases for either yet (correct me if I'm wrong)but I would like to discuss both in the more comprehensive context of the orchestration refactor.

If that sounds OK, we can create an issue with those 2 points under something like `Exception refactoring` and as related items come up they could be added as comments. Open to other ways of doing it though!

ehanson8:
And agreed on figuring this out however many commits it takes!

transmogrifier/exceptions.py
ghukill: Sounds good to me.  If `SkippedRecordEvent` is updated now to allow for default-y Exception behavior discussed above, then the rest feels like something we could suss out in future work.  There aren't that many instances of raising these from code, so would be easy to `ctrl + f` them and then think about what new exceptions might make sense to raise there.

ehanson8: New commit pushed!

jonavellecuerdo (approved):
~Looks good to me! Agree with the changes that were proposed by @ghukill.~ Still need to review the latest discussion/comments posted above.

transmogrifier/exceptions.py
```diff
@@ -0,0 +1,21 @@
...
+class SkippedRecordEvent(Exception):  # noqa: N818
+    """Exception raised for records that should be skipped.
+
+    Attributes:
+        source_record_id: The ID for the source record.
+    """
+
+    def __init__(self, message: str | None = None, source_record_id: str | None = None):
+        super().__init__(message)
+        self.source_record_id = source_record_id
```
ghukill, line 21: 👍

transmogrifier/sources/xml/datacite.py
```diff
@@ -54,7 +55,8 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
                 if self.valid_content_types([content_type]):
                     fields["content_type"] = [content_type]
                 else:
-                    return None
+                    message = f'Record skipped based on content type: "{content_type}"'
+                    raise SkippedRecordEvent(message, source_record_id)
```
ghukill, line 59: 👍 

I might suggest as we revist raising these in the future that we have a convention of positional `message` and then named arguments (e.g. `source_record_id=source_record_id`), but not blocking now as this definitely works too!

ghukill (approved):
Looks good!  Thanks for this intricate scaffolding work that is backwards/forwards compatible.  Think it's establishing a nice pattern for more granular exception raising as we progress.

transmogrifier/sources/xml/datacite.py
jonavellecuerdo: I like this idea, too!

jonavellecuerdo (approved):

transmogrifier/sources/xml/datacite.py
ehanson8: 👍

PR #177, opened 2024-05-21
Timx 283 oaidc field method refactor

jonavellecuerdo:

Field method refactor for base transform class `OaiDc`.

**Note:** Linters are failing for `transmogrifiers.sources.xml.springshare`, which will be addressed in a later PR!

1. Run `make test` and verify all tests are passing. 

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-283

### Developer

### Code Reviewer(s)


transmogrifier/sources/xml/oaidc.py
```diff
@@ -36,45 +36,34 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
         # citation: uses fallback get_citation() method
 
         # content_type
-        fields["content_type"] = [self.source]
+        fields["content_type"] = self.get_content_type()
```
jonavellecuerdo, line 39: I'd be interested in getting your thoughts on this approach. I opted to write a `get_*()` method  for consistency.

transmogrifier/sources/xml/oaidc.py
```diff
@@ -36,45 +36,34 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+        # [TODO] dates
+        fields["dates"] = self.get_dates(source_record_id, source_record) or None
 
         # edition: not set in this transformation
 
         # file_formats: not set in this transformation
 
         # format
-        fields["format"] = "electronic resource"
+        fields["format"] = self.get_format()
```
jonavellecuerdo, line 54: Similar to [the comment on content_type](https://github.com/MITLibraries/transmogrifier/pull/177/files#r1608711727), I opted to write a `get_*()` method for consistency, even though it only returns a hardcoded string.

tests/sources/xml/test_oai_dc.py
```diff
@@ -54,26 +65,217 @@ def test_oaidctransform_with_all_fields_transforms_correctly():
...
+def test_get_content_type_success(oai_dc_record_all_fields):
     source_records = OaiDc.parse_source_file(
-        f"{FIXTURES_PREFIX}/oaidc_record_valid_generic_date.xml"
+        "tests/fixtures/oai_dc/oaidc_record_all_fields.xml"
     )
-    transformer_instance = OaiDc("libguides", source_records)
-    xml = next(transformer_instance.source_records)
-    assert transformer_instance.get_dates("test_source_record_id", xml) == [
-        timdex.Date(kind="Unknown", note=None, range=None, value="2008-06-19T17:55:27")
+    assert OaiDc("cool-repo", source_records).get_content_type() == ["cool-repo"]
```
jonavellecuerdo, line 110: Because the `content_type` is derived from the `source` instance variable, `OaiDc` must be instantiated for the test.

tests/sources/xml/test_oai_dc.py
```diff
@@ -54,26 +65,217 @@ def test_oaidctransform_with_all_fields_transforms_correctly():
...
+    )
+
+
+def test_get_dates_transforms_correctly_if_date_invalid():
+    source_record = create_oai_dc_source_record_stub(
+        """
+        <dc:date>INVALID</dc:date>
+        """
+    )
+    assert OaiDc.get_dates(source_record_id="abc", source_record=source_record) == []
```
jonavellecuerdo, line 191: This test replaces the old test removed via [this commit](https://github.com/MITLibraries/transmogrifier/pull/177/commits/cd5569a419521205d9f95f8977bdfdc1b0dec2e8).

tests/sources/xml/test_oai_dc.py
```diff
@@ -54,26 +65,217 @@ def test_oaidctransform_with_all_fields_transforms_correctly():
...
+):
+    oai_dc_record_optional_fields_blank.header.identifier.clear()
+    assert OaiDc.get_identifiers(oai_dc_record_optional_fields_blank) == []
+
+
+def test_get_identifiers_transforms_correctly_if_fields_missing(
+    oai_dc_record_optional_fields_missing,
+):
+    oai_dc_record_optional_fields_missing.header.identifier.decompose()
+    assert OaiDc.get_identifiers(oai_dc_record_optional_fields_missing) == []
```
jonavellecuerdo, line 211: Curious what you think of this setup. 🤔 The identifiers are derived from a child of the `header` element, so using the stub (as it is written) or the `oai_dc_record_*` fixtures directly was not possible.

transmogrifier/sources/xml/oaidc.py
```diff
@@ -89,66 +78,102 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
         Method to get TIMDEX "dates" field.  This method broken out to allow subclasses
         to override.
 
         Return list of timdex.Date's if valid and present.
 
         Args:
-            source_record_id: Source record id
-            xml: A BeautifulSoup Tag representing a single OAI DC XML record.
+            source_record_id: Source record ID.
+            source_record : A BeautifulSoup Tag representing a single OAI DC record in XML.
```
ehanson8, line 114: Also remove the space before the colon here

tests/sources/xml/test_oai_dc.py
ehanson8: I think that works

ehanson8:
And let's resolve the `Signature of "get_dates"` linter issue before we merge this PR, happy to assist

tests/sources/xml/test_oai_dc.py
jonavellecuerdo: Just updated the hyperlink above. Thanks!

transmogrifier/sources/xml/springshare.py
```diff
@@ -99,8 +105,10 @@ def get_source_link(
...
+            source_record_id: Source record ID.
             source_base_url: Source base URL.
-            source_record_id: Record identifier for the source record.
-            xml: A BeautifulSoup Tag representing a single XML record.
+            source_record: A BeautifulSoup Tag representing a single OAI DC record in XML.
         """
-        return str(xml.find("dc:identifier").string)
+        if source_link := source_record.find("dc:identifier", string=True):
+            return str(source_link.string)
+        return ""
```
jonavellecuerdo, line 114: Previously, the method would throw an attribute error if `dc:identifier` did not appear in the XML: 

```
AttributeError: 'NoneType' object has no attribute 'string'
```

Given that `source_link` is a required field, raising an error seems to be the right approach, but is there a cleaner way to handle this case?

tests/sources/xml/test_oai_dc.py
```diff
@@ -1,12 +1,11 @@
 import pytest
-
 from bs4 import BeautifulSoup
 
 import transmogrifier.models as timdex
 from transmogrifier.sources.xml.oaidc import OaiDc
 
 
-def create_oai_dc_source_record_stub(xml_insert: str = "") -> BeautifulSoup:
+def create_oaidc_source_record_stub(xml_insert: str = "") -> BeautifulSoup:
```
ehanson8, line 8: I like the default, I'll update my with that

tests/conftest.py
```diff
@@ -111,6 +111,40 @@ def oai_pmh_records():
...
+    )
+    return next(source_records)
+
+
+@pytest.fixture
+def springshare_research_databases_record_optional_fields_missing():
+    source_records = OaiDc.parse_source_file(
+        "tests/fixtures/oai_dc/springshare/research_databases/research_databases_record_optional_fields_missing.xml"
+    )
+    return next(source_records)
```
ehanson8, line 145: As discussed in the `Datacite` PR, we can just use stub records instead of fixtures

tests/sources/xml/test_springshare.py
```diff
@@ -1,97 +1,179 @@
...
+    assert (
+        SpringshareOaiDc.get_dates(
+            source_record=source_record,
+            source_record_id="abc",
+        )
+        == []
+    )
+
+
+def test_get_dates_transforms_correctly_and_logs_error_if_date_invalid(
```
ehanson8, line 80: Perfect example of an edge case test!

transmogrifier/sources/xml/springshare.py
```diff
@@ -51,16 +52,16 @@ def get_dates(self, source_record_id: str, xml: Tag) -> list[timdex.Date]:
                 )
         return dates
```
ehanson8, line 53: As discussed in the other PR, this `return` can be updated with `or None`

transmogrifier/sources/xml/springshare.py
ehanson8: I think your instinct to use `SkippedRecordEvent` here is correct, @ghukill your thoughts?

tests/sources/xml/test_oai_dc.py
ghukill: +1

transmogrifier/sources/xml/oaidc.py
ghukill: I'm sure this has been hashed out in the comment above and elsewhere, but from my POV having these field methods is **key**.  When we get into orchestration, and it's not some imperative code like this where we can just set values, I'd imagine it will be looking for specifc methods to run, effectively making it the _only_ way to get values for those fields.

transmogrifier/sources/xml/springshare.py
ghukill: Agreed.  If we can't pull a `dc:identifier` value, then we don't have a source link to send people to, and it feels like it's not worth keeping the record (punting the awareness of this to later observability/metrics discussions).  

So yes, raising a `SkippedRecordEvent` feels like the right move!

tests/conftest.py
ghukill: Ya'll are deeper into these than I am, but agreed.  If we have pre-existing tests that use the full fixtures, seems okay to leave them (?), but any _new_ fixtures or tests ideally would be leaning on the stubs approach.

tests/sources/xml/test_oai_dc.py
ghukill: I had agreed with the comment until the strikethrough, lolz!  

See [this comment](https://github.com/MITLibraries/transmogrifier/pull/177/files#r1611753661) about how maybe keeping the function closer to the tests is handy, given we can modify the signature to set values as needed in the stub.  Maybe good fodder for our discussion today.

To me, feels like either approach is still kind of on the table.

transmogrifier/sources/xml/oaidc.py
```diff
@@ -36,45 +36,34 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
-                kind="OAI-PMH",
-            )
-        )
+        fields["identifiers"] = self.get_identifiers(source_record) or None
 
         # languages: not set in this transformation
 
         # links
-        fields["links"] = self.get_links(source_record_id, xml) or None
+        fields["links"] = self.get_links(source_record, source_record_id) or None
```
ghukill, line 66: Do we still need to pass `source_record_id`, if the `source_record` contains it?

And/or, could the method `get_links()` just call the _other_ method `get_source_record_id()` again if not?  

Feeling like maybe this is touched on in other comments.  Full disclaimer: I wrote these transforms fairly early on, and recall they had requirements that deviated from previous Transmog sources (as I think you've commented on as well @jonavellecuerdo, like the identifier vs source link).  So it feels possible that now is a good time to revisit some of those awkward edges.  Silver lining: these are not yet live, so I think we're fairly free to modify them somewhat heavily if needed.

ghukill (commented):
Submitting some comments now for our discussion, with the intent of completing a further review later.

tests/sources/xml/test_oai_dc.py
jonavellecuerdo: Update from @ghukill : Whole `header` element should be overwritten.

jonavellecuerdo:
@ghukill @ehanson8 Please see the latest three (3) commits for updates based on our sync yesterday! See [Summary of changes via Field Method Refactor
](https://docs.google.com/document/d/1JuKJRmmNK1tdrHYDzwEZf5ZTaChXnLwvJYi9G2nOfmo/edit#heading=h.6brephsq5mvh) for more details.

ghukill (commented):
Overall, thinking it's going well!  Appreciating that `OaiDc` and `Springshare` are kind of finicky, given there is heavy inheritance implications.

Left a couple of comments / questions.

tests/sources/xml/test_oai_dc.py
```diff
@@ -54,26 +66,331 @@ def test_oaidctransform_with_all_fields_transforms_correctly():
...
+            <identifier>oai:libguides.com:guides/175846</identifier>
+            """
+        ),
+        metadata_insert=(
+            """
+            <dc:date>2008-06-19T17:55:27</dc:date>
+            """
+        ),
+    )
+    assert OaiDc.get_dates(source_record=source_record) == [
```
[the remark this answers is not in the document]
jonavellecuerdo: @ghukill I'll follow this convention moving forward. I updated the function calls in tests to use positional arguments to match function signatures.

transmogrifier/sources/xml/springshare.py
```diff
@@ -49,18 +52,19 @@ def get_dates(self, source_record_id: str, xml: Tag) -> list[timdex.Date]:
                     source_record_id,
                     str(e),
                 )
-        return dates
+        return dates or None
 
-    def get_links(self, source_record_id: str, xml: Tag) -> list[timdex.Link] | None:
+    def get_links(self, source_record: Tag) -> list[timdex.Link] | None:
```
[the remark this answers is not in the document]
ghukill: @jonavellecuerdo - thanks for the explanation!  It's altogether possible that I introduced that when I first worked on these.  I feel like there is a bit of code smell there, and I'm probably to blame, but feels like maybe it's worth revisiting during the 2nd orchestration phase of this work.

ghukill (approved):
Looks good to me, thanks for the changes!

transmogrifier/sources/xml/springshare.py
```diff
@@ -49,18 +52,19 @@ def get_dates(self, source_record_id: str, xml: Tag) -> list[timdex.Date]:
...
         """
         Overrides OaiDc's default get_links() logic for Springshare records.
 
         Args:
-            source_record_id: Source record id
-            xml: A BeautifulSoup Tag representing a single OAI DC XML record.
+            source_record: A BeautifulSoup Tag representing a single OAI DC record in XML.
+            source_record_id: Source record ID.
```
ehanson8, line 63: No longer needed in the docstring since the param was removed

transmogrifier/sources/xml/springshare.py
```diff
@@ -31,11 +33,12 @@ def get_dates(self, source_record_id: str, xml: Tag) -> list[timdex.Date]:
         Additionally, only a single date will is expected.
 
         Args:
-            source_record_id: Source record id
-            xml: A BeautifulSoup Tag representing a single OAI DC XML record.
+            source_record: A BeautifulSoup Tag representing a single OAI DC record in XML.
+            source_record_id: Source record ID.
```
ehanson8, line 37: No longer needed in the docstring since the param was removed

transmogrifier/sources/xml/oaidc.py
```diff
@@ -89,87 +76,132 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
         Method to get TIMDEX "dates" field.  This method broken out to allow subclasses
         to override.
 
         Return list of timdex.Date's if valid and present.
 
         Args:
-            source_record_id: Source record id
-            xml: A BeautifulSoup Tag representing a single OAI DC XML record.
+            source_record: A BeautifulSoup Tag representing a single OAI DC record in XML.
+            source_record_id: Source record ID.
```
ehanson8, line 112: No longer needed in the docstring since the param was removed

ehanson8 (approved):
Approved, but remove references to `source_record_id` from the docstrings

PR #178, opened 2024-05-21
Add 1st set of Datacite field methods

ehanson8:
Refactors the first set of `Datacite` fields as separate field methods. As this is the first of many of these PRs that DataEng will be writing/reviewing, I encourage nit-picking so we can refine the overall approach early.

Run the following command to see that the `Datacite` transform still transforms a source file:

```
pipenv run transform -i tests/fixtures/datacite/datacite_records.xml -o output/datacite-transformed-records.json -s jpal
```

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-284

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_datacite.py
+++ b/tests/sources/xml/test_datacite.py
@@ -19,11 +19,14 @@ from transmogrifier.models import (
 from transmogrifier.sources.xml.datacite import Datacite
 
 
-def create_datacite_source_record_stub(xml_insert: str) -> BeautifulSoup:
+def create_datacite_source_record_stub(xml_insert: str = "") -> BeautifulSoup:
     xml_string = f"""
         <records>
          <record xmlns="http://www.openarchives.org/OAI/2.0/"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+          <header>
+            <identifier>abc123</identifier>
+          <header>
           <metadata>
            <resource xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xmlns="http://datacite.org/schema/kernel-4"
@@ -404,7 +407,7 @@ def test_get_alternate_titles_transforms_correctly_if_fields_blank():
 
 
 def test_get_alternate_titles_transforms_correctly_if_fields_missing():
-    source_record = create_datacite_source_record_stub("")
+    source_record = create_datacite_source_record_stub()
     assert Datacite.get_alternate_titles(source_record) is None
 
 
@@ -414,7 +417,7 @@ def test_get_content_type_success():
         <resourceType resourceTypeGeneral="Dataset">Survey Data</resourceType>
         """
     )
-    assert Datacite.get_content_type(source_record, "abc123") == ["Dataset"]
+    assert Datacite.get_content_type(source_record) == ["Dataset"]
 
 
 def test_get_content_type_transforms_correctly_if_fields_blank():
@@ -423,12 +426,12 @@ def test_get_content_type_transforms_correctly_if_fields_blank():
         <resourceType resourceTypeGeneral=""></resourceType>
         """
     )
-    assert Datacite.get_content_type(source_record, "abc123") is None
+    assert Datacite.get_content_type(source_record) is None
 
 
 def test_get_content_type_transforms_correctly_if_fields_missing():
-    source_record = create_datacite_source_record_stub("")
-    assert Datacite.get_content_type(source_record, "abc123") is None
+    source_record = create_datacite_source_record_stub()
+    assert Datacite.get_content_type(source_record) is None
 
 
 def test_get_contributors_success():
@@ -510,7 +513,7 @@ def test_get_contributors_transforms_correctly_if_fields_blank():
 
 
 def test_get_contributors_transforms_correctly_if_fields_missing():
-    source_record = create_datacite_source_record_stub("")
+    source_record = create_datacite_source_record_stub()
     assert Datacite.get_contributors(source_record) is None
 
 
--- a/tests/sources/xml/test_datacite.py
+++ b/tests/sources/xml/test_datacite.py
@@ -396,16 +396,16 @@ def test_get_alternate_titles_success():
     ]
 
 
-def test_get_alternate_titles_transforms_correctly_if_fields_blank(
-    datacite_record_optional_fields_blank,
-):
-    assert Datacite.get_alternate_titles(datacite_record_optional_fields_blank) == []
+def test_get_alternate_titles_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub(
+        '<titles><title titleType="AlternativeTitle"></title></titles>'
+    )
+    assert Datacite.get_alternate_titles(source_record) is None
 
 
-def test_get_alternate_titles_transforms_correctly_if_fields_missing(
-    datacite_record_optional_fields_missing,
-):
-    assert Datacite.get_alternate_titles(datacite_record_optional_fields_missing) == []
+def test_get_alternate_titles_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub("")
+    assert Datacite.get_alternate_titles(source_record) is None
 
 
 def test_get_content_type_success():
@@ -417,20 +417,18 @@ def test_get_content_type_success():
     assert Datacite.get_content_type(source_record, "abc123") == ["Dataset"]
 
 
-def test_get_content_type_transforms_correctly_if_fields_blank(
-    datacite_record_optional_fields_blank,
-):
-    assert (
-        Datacite.get_content_type(datacite_record_optional_fields_blank, "abc123") == []
+def test_get_content_type_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub(
+        """
+        <resourceType resourceTypeGeneral=""></resourceType>
+        """
     )
+    assert Datacite.get_content_type(source_record, "abc123") is None
 
 
-def test_get_content_type_transforms_correctly_if_fields_missing(
-    datacite_record_optional_fields_missing,
-):
-    assert (
-        Datacite.get_content_type(datacite_record_optional_fields_missing, "abc123") == []
-    )
+def test_get_content_type_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub("")
+    assert Datacite.get_content_type(source_record, "abc123") is None
 
 
 def test_get_contributors_success():
@@ -498,16 +496,22 @@ def test_get_contributors_success():
     ]
 
 
-def test_get_contributors_transforms_correctly_if_fields_blank(
-    datacite_record_optional_fields_blank,
-):
-    assert Datacite.get_contributors(datacite_record_optional_fields_blank) == []
+def test_get_contributors_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub(
+        """
+        <creators>
+         <creator />
+        <contributors>
+         <contributor />
+        </contributors>
+        """
+    )
+    assert Datacite.get_contributors(source_record) is None
 
 
-def test_get_contributors_transforms_correctly_if_fields_missing(
-    datacite_record_optional_fields_missing,
-):
-    assert Datacite.get_contributors(datacite_record_optional_fields_missing) == []
+def test_get_contributors_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub("")
+    assert Datacite.get_contributors(source_record) is None
 
 
 def test_generate_name_identifier_url_orcid_scheme(datacite_record_all_fields):
--- a/tests/sources/xml/test_datacite.py
+++ b/tests/sources/xml/test_datacite.py
@@ -1,3 +1,5 @@
+from bs4 import BeautifulSoup
+
 from transmogrifier.models import (
     AlternateTitle,
     Contributor,
@@ -15,6 +17,25 @@ from transmogrifier.models import (
     TimdexRecord,
 )
 from transmogrifier.sources.xml.datacite import Datacite
+
+
+def create_datacite_source_record_stub(xml_insert: str) -> BeautifulSoup:
+    xml_string = f"""
+        <records>
+         <record xmlns="http://www.openarchives.org/OAI/2.0/"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+          <metadata>
+           <resource xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xmlns="http://datacite.org/schema/kernel-4"
+           xsi:schemaLocation="http://datacite.org/schema/kernel-4
+           http://schema.datacite.org/meta/kernel-4.1/metadata.xsd">
+           {xml_insert}
+           </resource>
+          </metadata>
+         </record>
+        </records>
+        """
+    return BeautifulSoup(xml_string, "xml")
 
 
 def test_datacite_transform_with_all_fields_transforms_correctly(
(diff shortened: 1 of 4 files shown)


tests/sources/xml/test_datacite.py
```diff
@@ -359,6 +380,136 @@ def test_datacite_with_attribute_and_subfield_variations_transforms_correctly():
...
+    )
+    assert Datacite.get_content_type(source_record, "abc123") == ["Dataset"]
+
+
+def test_get_content_type_transforms_correctly_if_fields_blank(
+    datacite_record_optional_fields_blank,
+):
+    assert (
+        Datacite.get_content_type(datacite_record_optional_fields_blank, "abc123") == []
+    )
```
ghukill, line 425: Curious what would happen if we still used the stub, but passed it nothing?  Would the effect be the same?  Again, not a request, but per the PR notes, just some high level thinking and questions in this early PR.

ghukill (commented):
Overall, I think it looks great!

Left a few comments, but mostly just open questions and musings given this is one of the first PRs in this series.  

I'd likely just outright approve with my questions for discussion, but I was curious about one particular comment about the `self.get_<field_method>() or None` pattern,  pondering if it might be a good practice early here to make the field methods fully encapsulated to provide the final value.

transmogrifier/sources/xml/datacite.py
```diff
@@ -302,20 +248,113 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+                        for name_identifier in creator.find_all(
+                            "nameIdentifier", string=True
+                        )
+                    ]
+                    or None,
+                    kind="Creator",
+                )
+
+    @classmethod
+    def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
```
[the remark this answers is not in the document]
ehanson8: Would a docstring help here to clarify it's accessing the source record contributor?

tests/sources/xml/test_datacite.py
```diff
@@ -359,6 +380,136 @@ def test_datacite_with_attribute_and_subfield_variations_transforms_correctly():
...
+
+def test_get_content_type_transforms_correctly_if_fields_missing(
+    datacite_record_optional_fields_missing,
+):
+    assert (
+        Datacite.get_content_type(datacite_record_optional_fields_missing, "abc123") == []
+    )
+
+
+def test_get_contributors_success():
```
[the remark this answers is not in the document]
ehanson8: I appreciate that the test is fully encapsulated so even if it's big, it's all there rather than having to jump to the fixtures. So no complaints from me!

tests/sources/xml/test_datacite.py
ehanson8: A good suggestion, `blank` would need something passed but `missing` wouldn't, that would save us some fixtures in `conftest.py`. If @jonavellecuerdo agrees with that approach, I think I'll make the change

tests/sources/xml/test_datacite.py
```diff
@@ -17,6 +19,25 @@
...
+           xmlns="http://datacite.org/schema/kernel-4"
+           xsi:schemaLocation="http://datacite.org/schema/kernel-4
+           http://schema.datacite.org/meta/kernel-4.1/metadata.xsd">
+           {xml_insert}
+           </resource>
+          </metadata>
+         </record>
+        </records>
+        """
+    return BeautifulSoup(xml_string, "xml")
```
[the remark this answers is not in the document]
ehanson8: Great suggestion and that feels like an efficiency that would serve us better earlier in the refactor than later, @jonavellecuerdo your thoughts?

transmogrifier/sources/xml/datacite.py
```diff
@@ -13,100 +14,34 @@
...
-                    message = f'Record skipped based on content type: "{content_type}"'
-                    raise SkippedRecordEvent(message, source_record_id)
-        else:
-            logger.warning(
-                "Datacite record %s missing required Datacite field resourceType",
-                source_record_id,
-            )
+        fields["content_type"] = (
+            self.get_content_type(source_record, source_record_id) or None
+        )
```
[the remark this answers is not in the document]
ehanson8: Interesting, I see the logic in that but I feel that would be harder to pull out if (more likely when) it's deemed no longer necessary. A find-replace for ` or None` vs. removing:
```
 `if not xxxxx`: 
      return None
```
But curious on @jonavellecuerdo 's thoughts, I can be persuaded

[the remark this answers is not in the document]
ghukill: I feel like it'd be a small change in any given field method, but would have a larger effect higher up on the field method calling context where they would be totally uniform.

Taking this one as an example:
```python
@classmethod
def get_content_type(cls, source_record: Tag, source_record_id: str) -> list[str]:
    content_types = []
    if resource_type := source_record.metadata.find("resourceType"):
        if content_type := resource_type.get("resourceTypeGeneral"):
            if cls.valid_content_types([content_type]):
                content_types.append(str(content_type))
            else:
                message = f'Record skipped based on content type: "{content_type}"'
                raise SkippedRecordEvent(message, source_record_id)
    else:
        logger.warning(
            "Datacite record %s missing required Datacite field resourceType",
            source_record_id,
        )
    return content_types or None  #  <-------- might be just adding "or None" here?
```

Granted it would vary slightly for each, I also think the friction of thinking deeply about the return types and scenarios would also have benefits.  Feels like these methods should be _the_ value -- whatever it is -- that gets set on the `TIMDEXRecord`.  Now how we handle defualts there... that's another story, but at least we'd have more uniform methods to work with.

transmogrifier/sources/xml/datacite.py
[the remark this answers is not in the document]
ghukill: It's a good thought, but I don't think so.  

I would expect patterns like `get_dates()` calls `_get_published_dates()`, `get_subject_dates()`, etc.  This one is tricky because it's quite literally "contributors" in the datacite XML, so I think it's accurate.

Probably just a convention / norm, where these private helpers really aren't meant to be quickly scanned, but dug into when understanding a higher level field method.

transmogrifier/sources/xml/datacite.py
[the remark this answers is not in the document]
ehanson8: Ugh, you're totally right that's an easy shift, I withdraw my objection but still curious if @jonavellecuerdo has concerns

tests/sources/xml/test_datacite.py
jonavellecuerdo: This approach sounds good to me! I used this [method in updates to `SpringshareOaiDc` unit tests](https://github.com/MITLibraries/transmogrifier/pull/177/files#diff-9b1d193a2a05c09280b7b741221b67f2431bc44c9b4c69c3f79f09357314e8a3R131-R138) as well!

tests/sources/xml/test_datacite.py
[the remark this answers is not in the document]
jonavellecuerdo: I like the idea!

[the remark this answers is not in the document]
ehanson8: I'll create a ticket and I would propose one of us pick it up soon to save future refactoring

transmogrifier/sources/xml/datacite.py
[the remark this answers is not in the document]
jonavellecuerdo: No concerns. I like the idea of the field methods returning "_the_ value -- whatever it is -- that gets set on the `TIMDEXRecord`". I also think that would align more closely with [the type hints set for the fields in the `TimdexRecord` model](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/models.py#L159-L215).

ehanson8:
@ghukill @jonavellecuerdo Pushed a new commit with the changes we discussed

ehanson8:
@ghukill @jonavellecuerdo Another update after our very helpful discussion this morning!

transmogrifier/sources/xml/datacite.py
```diff
@@ -302,20 +246,117 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+        return alternate_titles or None
+
+    @classmethod
+    def _get_additional_titles(
+        cls, source_record: Tag
+    ) -> Iterator[timdex.AlternateTitle]:
+        """Get additional titles from get_main_titles method."""
+        for index, title in enumerate(cls.get_main_titles(source_record)):
+            if index > 0:
+                yield timdex.AlternateTitle(value=title)
```
jonavellecuerdo, line 273: Hmm, the creation of the private method `_get_additional_titles()` brings the number of `title`-related methods to 4: 
* `get_valid_title`
* `get_main_titles`
* `get_alternate_titles`
* `_get_additional_titles`

Because of this, I personally prefer moving the logic in `_get_additional_titles()` private method into `get_alternate_titles()` and adding a comment instead. If not, maybe rename to `_get_additional_alternate_titles()`. 

That said, I think we did kinda' discuss that `get_valid_title()` might be revisited at some point...

Curious what @ghukill thinks!

transmogrifier/sources/xml/datacite.py
```diff
@@ -302,20 +246,117 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+                else:
+                    message = f'Record skipped based on content type: "{content_type}"'
+                    raise SkippedRecordEvent(
+                        message, cls.get_source_record_id(source_record)
+                    )
+        else:
+            logger.warning(
+                "Datacite record %s missing required Datacite field resourceType",
+                cls.get_source_record_id(source_record),
+            )
```
jonavellecuerdo, line 291: Given that `content_type` is not a required field, I would opt to remove lines 287-291.  The warning is confusing as it is technically "not required".  This might've been related to our discussion earlier about "useful logging"? 🤔

transmogrifier/sources/xml/datacite.py
```diff
@@ -302,20 +246,117 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+            if index > 0:
+                yield timdex.AlternateTitle(value=title)
+
+    @classmethod
+    def get_content_type(cls, source_record: Tag) -> list[str] | None:
+        content_types = []
+        if resource_type := source_record.metadata.find("resourceType"):
+            if content_type := resource_type.get("resourceTypeGeneral"):
+                if cls.valid_content_types([content_type]):
+                    content_types.append(str(content_type))
```
jonavellecuerdo, line 281: Small update but perhaps rewrite to: 

```
content_types.append(str(content_type)
if cls.valid_content_types(content_types):
    return content_types
else: 
    <...>
```
This would remove the use of `or None` but might be a smoother read?

jonavellecuerdo (commented):
@ehanson8 Just have a couple of questions and suggestions for changes. Let me know what you think! The `Datacite` transform has some complex logic, and I appreciate your efforts to make these updates as clear and digestible as possible. :)

transmogrifier/sources/xml/datacite.py
```diff
@@ -261,7 +261,7 @@ def get_alternate_titles(cls, source_record: Tag) -> list[timdex.AlternateTitle]
             ]
         )
         alternate_titles.extend(list(cls._get_additional_titles(source_record)))
-        return alternate_titles
+        return alternate_titles or None
```
ghukill, line 264: At this time, I do like this full encapsulation, thanks for making these changes.

Relevant here and elsewhere...

Once we get into orchestration, which I think touches on the instantiation of the `TIMDEXRecord` instance, it might be worth considering if maybe it's okay to return an empty list?  If so, and that was filtered out during serialization to JSON, then we could probably drop a bunch of these conversions to `None`.  Just noting for the future, but again, think this is great for now, and in the spirit of the incremental approach.

ghukill (approved):
Looks good to me!

transmogrifier/sources/xml/datacite.py
ghukill: Thanks for raising @jonavellecuerdo.  Without comment on your proposal above, I see something else in these title parsings that also stood out, after some hopping back and forth.

Looks like maybe the same XML parsing is used [here in get_main_titles()](https://github.com/MITLibraries/transmogrifier/blob/bdb2c4bd95c930683f7a5b6c2499fc88fe03eebb/transmogrifier/sources/xml/datacite.py#L357-L359) and then again [here in get_alternate_titles()](https://github.com/MITLibraries/transmogrifier/blob/bdb2c4bd95c930683f7a5b6c2499fc88fe03eebb/transmogrifier/sources/xml/datacite.py#L255-L260)?  The difference being the first is looking for strings, and the second is returning `timdex.AlternateTitle` instances.

I don't know what the solution is, something is feeling a bit wonky here.  

Wouldn't this result in duplicating values for "main titles" and "alternate titles"?  Does that have some bearing here?

transmogrifier/sources/xml/datacite.py
```diff
@@ -302,20 +246,117 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+        alternate_titles.extend(
+            [
+                timdex.AlternateTitle(
+                    value=str(title.string.strip()),
+                    kind=title["titleType"],
+                )
+                for title in source_record.find_all("title", string=True)
+                if title.get("titleType")
+            ]
+        )
```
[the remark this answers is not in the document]
ehanson8: Good call, thanks!

transmogrifier/sources/xml/datacite.py
ehanson8: I agree this should go but propose we create an issue of "evaluate logging for usefulness"  to handle it globally (with other items flagged in the comments as they are encountered) and not bog down the field method refactor.  The origin of this is that `resourceType` is required by the `Datacite` schema, but we've never done anything with it and it's incongruous with how `transmogrifier` logging has developed since. `Datacite` was the first transform written and the context was the RDI project, which is no longer our context

transmogrifier/sources/xml/datacite.py
ehanson8: I am most likely onboard with that approach when we get there!

transmogrifier/sources/xml/datacite.py
ehanson8: This is another one where I agree 100% but I think we're better off creating an issue to address this separately. This is getting too close to core logic that I'm hesitant to mess with it during the field method refactor. Agree or disagree with that approach?

jonavellecuerdo: @ghukill I think the difference between `get_main_titles()` and `get_alternate_titles` is:
* `get_main_titles()` retrieves values where attribute `@titleType` is None. 
* `get_alternate_titles()` retrieves values where attribute `@titleType` exists.
* `_get_additional_titles` retrieves all values from `get_main_titles()`, excluding the first item in the list (if there are multiple values returned for `get_main_titles()`, the first value is used for `timdexRecord.title`.

That said, @ehanson8 you make a good point! This is stepping into the logic as opposed to the field method. I started an [issue for Datacite transform logic](https://github.com/MITLibraries/transmogrifier/issues/179).

transmogrifier/sources/xml/datacite.py
jonavellecuerdo: Ah, thank you for clarifying! Created an [issue](https://github.com/MITLibraries/transmogrifier/issues/180) for future work and discussion. :)

jonavellecuerdo (approved):

transmogrifier/sources/xml/datacite.py
ghukill: Agreed @ehanson8, and thanks for the eagle eyes @jonavellecuerdo!  This is like the 100th time this week I've missed a `not` in code.  I might need new glasses. 

Thanks for starting the issue.  It does feel like something that could get untangled and streamlined, but agree with all comments above.

transmogrifier/sources/xml/datacite.py
ehanson8: Thank you for creating the issue!

transmogrifier/sources/xml/datacite.py
ehanson8: Thanks again for creating the issue, though I would argue that `title` issue goes much deeper than just `Datacite` so you could frame it more broadly as `Refactor title logic`. I think other transforms may be equally messy re: titles

Issue #180, opened 2024-05-24
Evaluate logging for usefulness


jonavellecuerdo:
Scope improvements to logging

* @jonavellecuerdo 's [comment on `Datacite` transform](https://github.com/MITLibraries/transmogrifier/pull/178/files/bdb2c4bd95c930683f7a5b6c2499fc88fe03eebb#r1612244336) on PR #178

ehanson8:
Also evaluate the `Datacite` `publicationYear` [warning](https://github.com/MITLibraries/transmogrifier/pull/181#discussion_r1619352558), `publisher`, and `@descriptionType` warnings

PR #181, opened 2024-05-29
Add 2nd set of Datacite field methods

ehanson8:
Continue refactoring `Datacite` transform to use field methods.

Run the following command to see that the `Datacite` transform still transforms a source file:

```
pipenv run transform -i tests/fixtures/datacite/datacite_records.xml -o output/datacite-transformed-records.json -s jpal
```

NO

NO

* https://mitlibraries.atlassian.net/browse/TIMX-284

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_datacite.py
+++ b/tests/sources/xml/test_datacite.py
@@ -517,6 +517,246 @@ def test_get_contributors_transforms_correctly_if_fields_missing():
     assert Datacite.get_contributors(source_record) is None
 
 
+def test_get_dates_success():
+    source_record = create_datacite_source_record_stub(
+        """
+        <publicationYear>2017</publicationYear>
+        <dates>
+         <date dateType="Submitted">2017-02-27</date>
+         <date dateType="Updated"
+         dateInformation="This was updated on this date">2019-06-24</date>
+         <date dateType="Collected">2007-01-01/2007-02-28</date>
+        </dates>
+        """
+    )
+    assert Datacite.get_dates(source_record) == [
+        Date(kind="Publication date", value="2017"),
+        Date(kind="Submitted", value="2017-02-27"),
+        Date(kind="Updated", note="This was updated on this date", value="2019-06-24"),
+        Date(
+            kind="Collected",
+            range=DateRange(gte="2007-01-01", lte="2007-02-28"),
+        ),
+    ]
+
+
+def test_get_dates_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub(
+        """
+        <publicationYear />
+        <dates>
+         <date />
+        </dates>
+        """
+    )
+    assert Datacite.get_dates(source_record) is None
+
+
+def test_get_dates_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub()
+    assert Datacite.get_dates(source_record) is None
+
+
+def test_get_edition_success():
+    source_record = create_datacite_source_record_stub("<version>1.2</version>")
+    assert Datacite.get_edition(source_record) == "1.2"
+
+
+def test_get_edition_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub("<version />")
+    assert Datacite.get_edition(source_record) is None
+
+
+def test_get_edition_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub()
+    assert Datacite.get_edition(source_record) is None
+
+
+def test_get_file_formats_success():
+    source_record = create_datacite_source_record_stub(
+        """
+        <formats>
+         <format>application/vnd.openxmlformats-officedocument.spreadsheetml.sheet</format>
+         <format>application/pdf</format>
+         <format>application/pdf</format>
+         <format>application/vnd.openxmlformats-officedocument.spreadsheetml.sheet</format>
+         <format>application/pdf</format>
+         <format>application/x-stata-syntax</format>
+         <format>application/x-stata</format>
+         <format>application/x-stata</format>
+         <format>application/zip</format>
+         <format>application/pdf</format>
+         <format>application/pdf</format>
+        </formats>
+        """
+    )
+    assert Datacite.get_file_formats(source_record) == [
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        "application/pdf",
+        "application/pdf",
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        "application/pdf",
+        "application/x-stata-syntax",
+        "application/x-stata",
+        "application/x-stata",
+        "application/zip",
+        "application/pdf",
+        "application/pdf",
+    ]
+
+
+def test_get_file_formats_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub("<formats><format /></formats>")
+    assert Datacite.get_file_formats(source_record) is None
+
+
+def test_get_file_formats_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub()
+    assert Datacite.get_file_formats(source_record) is None
+
+
+def test_get_format_success():
+    assert Datacite.get_format() == "electronic resource"
+
+
+def test_get_funding_information_success():
+    source_record = create_datacite_source_record_stub(
+        """
+        <fundingReferences>
+         <fundingReference>
+          <funderName>3ie, Nike Foundation</funderName>
+          <funderIdentifier
+          funderIdentifierType="Crossref FunderID">0987</funderIdentifier>
+          <awardNumber awardURI="http://awards.example/7689">OW1/1012 (3ie)</awardNumber>
+         </fundingReference>
+        </fundingReferences>
+        """
+    )
+    assert Datacite.get_funding_information(source_record) == [
+        Funder(
+            funder_name="3ie, Nike Foundation",
+            funder_identifier="0987",
+            funder_identifier_type="Crossref FunderID",
+            award_number="OW1/1012 (3ie)",
+            award_uri="http://awards.example/7689",
+        )
+    ]
+
+
+def test_get_funding_information_transforms_correctly_if_fields_blank():
+    source_record = create_datacite_source_record_stub(
+        "<fundingReferences><fundingReference /></fundingReferences>"
+    )
+    assert Datacite.get_funding_information(source_record) is None
+
+
+def test_get_funding_information_transforms_correctly_if_fields_missing():
+    source_record = create_datacite_source_record_stub()
+    assert Datacite.get_funding_information(source_record) is None
+
+
+def test_get_identifiers_success():
+    source_record = create_datacite_source_record_stub(
+        """
+        <identifier identifierType="DOI">10.7910/DVN/19PPE7</identifier>
+        <alternateIdentifiers>
+         <alternateIdentifier alternateIdentifierType="url">https://zenodo.org/record/5524465</alternateIdentifier>
+        </alternateIdentifiers>
+        <relatedIdentifiers>
+         <relatedIdentifier relatedIdentifierType="DOI" relationType="IsCitedBy">
+         10.1257/app.20150390</relatedIdentifier>
+         <relatedIdentifier relationType="IsVersionOf">10.5281/zenodo.5524464
+         </relatedIdentifier>
+         <relatedIdentifier relatedIdentifierType="ISBN"
+         relationType="IsIdenticalTo">1234567.5524464</relatedIdentifier>
+         <relatedIdentifier relatedIdentifierType="ISBN" relationType="Other">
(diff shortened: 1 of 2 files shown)


tests/sources/xml/test_datacite.py
```diff
@@ -517,6 +517,246 @@ def test_get_contributors_transforms_correctly_if_fields_missing():
...
+def test_get_links_success(datacite_record_all_fields):
+    source_record = create_datacite_source_record_stub()
+    datacite_transformer = Datacite("jpal", datacite_record_all_fields)
+    assert datacite_transformer.get_links(source_record) == [
+        Link(
+            url="https://dataverse.harvard.edu/dataset.xhtml?persistentId=abc123",
+            kind="Digital object URL",
+            text="Digital object URL",
+        )
+    ]
```
ehanson8, line 730: This came up in Jonavelle's PR, the code requires instantiation for `self.source_base_url`, let me know if there's a better approach here. Also because this is essentially just wrapping `get_source_id`, I felt any edge case tests should probably happen around that method. But let me know if you disagree

ghukill: FWIW, I think it's fine as-is.  I'd like to continue discussing when class vs normal methods make sense, but feel like that might be something we could investigate wholesale later, then revisit these edge cases.

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +238,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
...
+            if validate_date(
+                publication_year,
+                cls.get_source_record_id(source_record),
+            ):
+                yield timdex.Date(kind="Publication date", value=publication_year)
+        else:
+            logger.warning(
+                "Datacite record %s missing required Datacite field publicationYear",
+                cls.get_source_record_id(source_record),
+            )
```
ghukill, line 266: I see we log a warning, but this doesn't rise to the level of skipping the record?

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +238,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
...
+
+    @classmethod
+    def _get_dates(cls, source_record: Tag) -> Iterator[timdex.Date]:
```
ghukill, line 269: Forgive me if we've already discussed this, but I keep tripping on this private method `_get_dates()` and it's relationship to `get_dates()`.  What about a rename to something like `_get_other_dates()`?  I realize that "other" isn't very exact, but I think it definitively differentiates from the higher level field  method `get_dates()`.

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +238,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
                     kind=contributor.get("contributorType") or "Not specified",
                 )
 
+    @classmethod
+    def get_dates(
```
ghukill, line 242: Commenting high up here, but I continue to have the feeling the dates logic and handling as a whole could be refactored some.   But, I think it could wait.  It feels like there are tethers to the `validate_date()` which consistently complicate the logic, and depending on how that lands, probably isn't the right time to tackel.

FWIW, I _do_ think this is a nice port into the field method approach, which is the goal here.

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +238,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
...
+                    value=str(identifier_element.string),
+                    kind=identifier_element.get("identifierType") or "Not specified",
+                )
+            )
+        identifiers.extend(list(cls._get_alternateidentifiers(source_record)))
+        identifiers.extend(list(cls._get_relatedidentifiers(source_record)))
+        return identifiers or None
+
+    @classmethod
+    def _get_alternateidentifiers(
```
ghukill, line 366: Maybe another underscore, ` _get_alternate_identifiers()`?

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +238,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
...
+            "alternateIdentifier", string=True
+        ):
+            yield timdex.Identifier(
+                value=str(alternate_identifier_element.string),
+                kind=alternate_identifier_element.get("alternateIdentifierType")
+                or "Not specified",
+            )
+
+    @classmethod
+    def _get_relatedidentifiers(
```
ghukill, line 380: Same here, maybe ` _get_related_identifiers()`?

ghukill (commented):
Overall, looking real good.  Left a couple of comments, but nothing major.  I'm finding the file, each time, easier and easier to scan.

tests/sources/xml/test_datacite.py
ehanson8: Agreed that this should be revisited globally, I'll create an issue

tests/sources/xml/test_datacite.py
```diff
@@ -517,6 +517,246 @@ def test_get_contributors_transforms_correctly_if_fields_missing():
...
+        "application/pdf",
+        "application/pdf",
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        "application/pdf",
+        "application/x-stata-syntax",
+        "application/x-stata",
+        "application/x-stata",
+        "application/zip",
+        "application/pdf",
+        "application/pdf",
```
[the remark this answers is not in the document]
ehanson8: Alternative approach: what about creating a function to dedupe all fields after the source record has been processed? Similar to `create_dates_and_locations_from_publishers` and `create_locations_from_spatial_subjects`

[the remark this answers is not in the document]
ghukill: A thousand times yes.  Unless we have a very good reason to, it feels like we could dedupe _all_ TIMDEX fields that are a list.  

Actually, `attrs` may really help us out here.  And/or `set`.  There are many ways we could dedupe a field, even if they are complex `attrs` classes.

transmogrifier/sources/xml/datacite.py
ehanson8: This is another `Datacite`-related warning like what @jonavellecuerdo commented about [here](https://github.com/MITLibraries/transmogrifier/pull/178/files/bdb2c4bd95c930683f7a5b6c2499fc88fe03eebb#r1612244336). I'll add it to the [issue](https://github.com/MITLibraries/transmogrifier/issues/180) because it should also likely be removed.

transmogrifier/sources/xml/datacite.py
ehanson8: What about get `_get_datacite_dates` which I think is more accurate? I could also update `_get_contributors` to `_get_datacite_contributors` since that's what originally prompted the conversation

transmogrifier/sources/xml/datacite.py
ehanson8: Agreed that this is another issue better handled globally and later on

transmogrifier/sources/xml/datacite.py
ehanson8: I was going for the actual `Datacite` element name but you're correct that `_get_alternate_identifiers` is easier to read. I'll update

tests/sources/xml/test_datacite.py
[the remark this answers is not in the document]
ehanson8: Great, created an [issue](https://github.com/MITLibraries/transmogrifier/issues/184)

transmogrifier/sources/xml/datacite.py
ghukill: Ah thanks for the reminder, I thought it had come up.  

I don't mean to be difficult, but even adding `datacite` feels odd... like we know it's datacite already.   But I don't really have a better solution.  

I suppose one unique property is that we're setting the date `kind` dynamically by the attribute `dateType`; wonder if that could come into play naming wise?  Maybe something like `_get_dates_by_datetype_attribute()`?  It's _long_... but it tells you precisely what's going on?

Or we could focus on the element used, like `_get_dates_by_date_element()`?  

Open to pushback here.  If folks are okay with `_get_dates()` and `_get_contributors()` I could be too.

ehanson8: I like `_get_dates_by_date_element`, I think I'll go with that

ehanson8:
@ghukill @jonavellecuerdo Pushed a new commit based on Graham's comments

ghukill (approved):
I dig it!  Thanks for the changes.  At this stage, they feel really understandable.  Maybe in "phase 2" we take another pass at naming, once we've seen a lot, but appreciate the changes now.

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +242,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
...
+                value=cls.generate_related_item_identifier_url(
+                    related_identifier_element
+                ),
+                kind=str(related_identifier_element["relationType"]),
+            )
+
+    @classmethod
+    def get_languages(cls, source_record: Tag) -> list[str] | None:
+        languages = []
+        if language := source_record.metadata.find("language", string=True):
```
jonavellecuerdo, line 406: [Non-blocking] When do we include the "_element" suffix to a variable?

* Some cases, we just use the field name:
   * [get_funding_information ](https://github.com/MITLibraries/transmogrifier/pull/181/files#diff-875ef9d9184fe0a891a9abf429a0b3187d70ef66c9908a9e0a4008cb26995afdR334)
   * get_languages
   * [get_edition](https://github.com/MITLibraries/transmogrifier/pull/181/files#diff-875ef9d9184fe0a891a9abf429a0b3187d70ef66c9908a9e0a4008cb26995afdR316)
* Other cases, we include the suffix "_element": 
   * [get_identifier](https://github.com/MITLibraries/transmogrifier/pull/181/files#diff-875ef9d9184fe0a891a9abf429a0b3187d70ef66c9908a9e0a4008cb26995afdR358) 
   * [_get_alternate_identifier](https://github.com/MITLibraries/transmogrifier/pull/181/files#diff-875ef9d9184fe0a891a9abf429a0b3187d70ef66c9908a9e0a4008cb26995afdR374-R376)

This is something I've personally been getting tripped up on in the transforms I've been working on --naming conventions assigned to `find`/`find_all` method calls. These differences don't cause any problems but consistency would be nice...

@ghukill @ehanson8

jonavellecuerdo (commented):
One small change request and a question for both of you (@ghukill and @ehanson8).

transmogrifier/sources/xml/datacite.py
ehanson8: I was using the `_element` convention where element attributes are accessed in addition to `element.string` so it probably should be used for `get_funding_information` as well. To me, it doesn't feel as necessary when we're just grabbing `element.string`. Interested in consistency as well so I want to hear everyone's thoughts

transmogrifier/sources/xml/datacite.py
```diff
@@ -339,6 +242,189 @@ def _get_contributors(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
...
+
+    @classmethod
+    def get_funding_information(cls, source_record: Tag) -> list[timdex.Funder] | None:
+        funding_information = []
+        for funding_reference in source_record.metadata.find_all("fundingReference"):
+            funder = timdex.Funder()
+            if funder_name := funding_reference.find("funderName", string=True):
+                funder.funder_name = str(funder_name.string)
+            if award_number := funding_reference.find("awardNumber"):
+                funder.award_number = award_number.string or None
```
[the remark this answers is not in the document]
ehanson8: I should have made a note about this but forgot, `str(award_number.string)` returns `"None"` if there is no `.string`. However, we can't do `funding_reference.find("awardNumber", string=True)` because we're also interested in the `awardURI` attribute even if there isn't `award_number.string`. Not sure if there is a better way to handle this but decided to make a one-off exception that rule, happy to update if anyone has suggestions!

transmogrifier/sources/xml/datacite.py
ghukill: I see the thought process there @ehanson8, and I agree that consistency would be nice, but it doesn't feel like a deal breaker to me with some variation.  Given how common the convention is to a) find an element, and then b) use that element variable to extract a string _or_ an attribute value, I think the `_element` could get pretty chatty.  

It might be worth an issue?  Then we could keep moving through this refactor, but make a global pass at some point.  

In summary: I'm personally okay with variations, given different times in codebae, but an issue to revist could be nice.

jonavellecuerdo: Thanks for sharing! I'll apply that thought process as I work on other transforms, @ehanson8 .

jonavellecuerdo (approved):

PR #185, opened 2024-05-30
Timx 287 ead field method refactor 1

jonavellecuerdo:

Field method refactor for transform class Ead (Part 1).

* Added private"methods for key elements in EAD-formatted XML records (`<archdesc>`, `<archdesc><did>`, `<archdesc><controlaccess>`
* Added field methods and corresponding unit tests for `alternate_titles`, `citation`, `content_type`

**Note:** For the time-being, I've added calls to the private methods at the beginning of `get_optional_fields()`, which is required by field derivations that **will eventually be moved into their own field methods**.

2. Run CLI command
   ```
   pipenv run transform -i tests/fixtures/ead/ead_record_all_fields.xml -o output/ead-transformed-records.json -s aspace
   ```
   Output: 
   ```
   2024-05-30 12:33:59,914 INFO transmogrifier.cli.main(): Logger 'root' configured with level=INFO
   2024-05-30 12:33:59,914 INFO transmogrifier.cli.main(): No Sentry DSN found, exceptions will not be sent to Sentry
   2024-05-30 12:33:59,914 INFO transmogrifier.cli.main(): Running transform for source aspace
   2024-05-30 12:33:59,930 WARNING transmogrifier.sources.transformer.get_valid_title(): Record repositories/2/resources/1 has multiple titles. Using the first title from the following titles found: ['Charles J. Connick Stained Glass Foundation Collection', 'Title 2', 'Title 3']
   2024-05-30 12:33:59,930 INFO transmogrifier.cli.main(): Completed transform, total records processed: 1, transformed records: 1, skipped records: 0, deleted records: 0
   2024-05-30 12:33:59,930 INFO transmogrifier.cli.main(): Total time to complete transform: 0:00:00.016147

   ```
NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-287

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_ead.py
+++ b/tests/sources/xml/test_ead.py
@@ -1,10 +1,80 @@
 import logging
+from typing import Literal
+
+from bs4 import BeautifulSoup
 
 import transmogrifier.models as timdex
 from transmogrifier.sources.xml.ead import Ead
 
 
-def test_ead_record_all_fields_transform_correctly():
+def create_ead_source_record_stub(
+    header_insert: str = "",
+    metadata_insert: str = "",
+    parent_element: Literal["archdesc", "did"] = None,  # noqa: RUF013
+) -> BeautifulSoup:
+    """
+    Create source record for unit tests.
+
+    Args:
+        header_insert (str): For EAD-formatted XML, the <header> element is used
+            in the derivation of 'source_record_id'.
+        metadata_insert (str): An string representing a metadata XML element.
+        parent_element (Literal["archdesc", "did"]): For EAD-formatted XML,
+            all of the information about a collection that is relevant to the
+            TIMDEX data model is encapsulated within the <archdesc level="collection">
+            element. Metadata will be formatted either as a direct descendant of
+            the <archdesc> element or a descendant of the <did> element.
+
+    Note: A source record for "missing" field method tests can be created by
+        leaving metadata_insert = "" (the default) and setting parent_element.
+    """
+    xml_string = """
+        <records>
+            <record xmlns="http://www.openarchives.org/OAI/2.0/"
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+                <header>
+                    {header_insert}
+                </header>
+                <metadata>
+                    <ead xmlns="urn:isbn:1-931666-22-9" xmlns:xlink="http://www.w3.org/1999/xlink"
+                        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                        xsi:schemaLocation="urn:isbn:1-931666-22-9 http://www.loc.gov/ead/ead.xsd">
+                        <eadheader countryencoding="iso3166-1" dateencoding="iso8601"
+                            findaidstatus="completed"
+                            langencoding="iso639-2b"
+                            repositoryencoding="iso15511">
+                            <eadid countrycode="US" mainagencycode="US-mcm">
+                                VC-0002
+                            </eadid>
+                        </eadheader>
+                        {metadata_insert}
+                    </ead>
+                </metadata>
+            </record>
+        </records>
+    """
+    if metadata_insert and parent_element is None:
+        message = (
+            "Argument 'parent_element' cannot be of NoneType "
+            "if 'metadata_insert' is set."
+        )
+        raise TypeError(message)
+    if parent_element == "archdesc":
+        _metadata_insert = f"""
+            <archdesc level="collection">{metadata_insert}</archdesc>
+            """
+    elif parent_element == "did":
+        _metadata_insert = f"""
+            <archdesc level="collection"><did>{metadata_insert}</did></archdesc>
+            """
+
+    return BeautifulSoup(
+        xml_string.format(header_insert=header_insert, metadata_insert=_metadata_insert),
+        "xml",
+    )
+
+
+def test_ead_transform_with_all_fields_transforms_correctly():
     ead_xml_records = Ead.parse_source_file(
         "tests/fixtures/ead/ead_record_all_fields.xml"
     )
@@ -218,35 +288,43 @@ def test_ead_record_all_fields_transform_correctly():
     )
 
 
-def test_ead_record_with_missing_archdesc_logs_error(caplog):
+def test_ead_transform_with_optional_fields_blank_transforms_correctly():
     ead_xml_records = Ead.parse_source_file(
-        "tests/fixtures/ead/ead_record_missing_archdesc.xml"
+        "tests/fixtures/ead/ead_record_blank_optional_fields.xml"
     )
     output_records = Ead("aspace", ead_xml_records)
-    assert len(list(output_records)) == 0
-    assert output_records.processed_record_count == 1
-    assert (
-        "transmogrifier.sources.xml.ead",
-        logging.ERROR,
-        "Record ID repositories/2/resources/4 is missing archdesc element",
-    ) in caplog.record_tuples
+    assert next(output_records) == timdex.TimdexRecord(
+        source="MIT ArchivesSpace",
+        source_link="https://archivesspace.mit.edu/repositories/2/resources/2",
+        timdex_record_id="aspace:repositories-2-resources-2",
+        title="Title not provided",
+        citation=(
+            "Title not provided. Archival materials. "
+            "https://archivesspace.mit.edu/repositories/2/resources/2"
+        ),
+        content_type=["Archival materials"],
+    )
 
 
-def test_ead_record_with_missing_archdesc_did_logs_error(caplog):
+def test_ead_transform_with_optional_fields_missing_transforms_correctly():
     ead_xml_records = Ead.parse_source_file(
-        "tests/fixtures/ead/ead_record_missing_archdesc_did.xml"
+        "tests/fixtures/ead/ead_record_missing_optional_fields.xml"
     )
     output_records = Ead("aspace", ead_xml_records)
-    assert len(list(output_records)) == 0
-    assert output_records.processed_record_count == 1
-    assert (
-        "transmogrifier.sources.xml.ead",
-        logging.ERROR,
-        "Record ID repositories/2/resources/3 is missing archdesc > did element",
-    ) in caplog.record_tuples
+    assert next(output_records) == timdex.TimdexRecord(
+        source="MIT ArchivesSpace",
+        source_link="https://archivesspace.mit.edu/repositories/2/resources/5",
+        timdex_record_id="aspace:repositories-2-resources-5",
+        title="Title not provided",
+        citation=(
+            "Title not provided. Archival materials. "
+            "https://archivesspace.mit.edu/repositories/2/resources/5"
+        ),
+        content_type=["Archival materials"],
+    )
 
 
-def test_ead_record_with_attribute_and_subfield_variations_transforms_correctly():
+def test_ead_transform_with_attribute_and_subfield_variations_transforms_correctly():
     ead_xml_records = Ead.parse_source_file(
         "tests/fixtures/ead/ead_record_attribute_and_subfield_variations.xml"
     )
@@ -468,25 +546,28 @@ def test_ead_record_with_attribute_and_subfield_variations_transforms_correctly(
     )
 
 
-def test_ead_record_with_blank_optional_fields_transforms_correctly():
+def test_ead_transform_with_missing_archdesc_skips_record():
     ead_xml_records = Ead.parse_source_file(
-        "tests/fixtures/ead/ead_record_blank_optional_fields.xml"
+        "tests/fixtures/ead/ead_record_missing_archdesc.xml"
     )
     output_records = Ead("aspace", ead_xml_records)
-    assert next(output_records) == timdex.TimdexRecord(
-        source="MIT ArchivesSpace",
-        source_link="https://archivesspace.mit.edu/repositories/2/resources/2",
-        timdex_record_id="aspace:repositories-2-resources-2",
-        title="Title not provided",
-        citation=(
-            "Title not provided. Archival materials. "
-            "https://archivesspace.mit.edu/repositories/2/resources/2"
-        ),
-        content_type=["Archival materials"],
+    assert len(list(output_records)) == 0
+    assert output_records.processed_record_count == 1
+    assert output_records.skipped_record_count == 1
+
+
+def test_ead_transform_with_missing_archdesc_did_skips_record():
+    ead_xml_records = Ead.parse_source_file(
+        "tests/fixtures/ead/ead_record_missing_archdesc_did.xml"
     )
+    output_records = Ead("aspace", ead_xml_records)
+
+    assert len(list(output_records)) == 0
+    assert output_records.processed_record_count == 1
+    assert output_records.skipped_record_count == 1
 
 
-def test_ead_record_invalid_date_and_date_range_are_omitted(caplog):
+def test_ead_transform_with_invalid_date_and_date_range_omits_dates(caplog):
     caplog.set_level(logging.DEBUG)
     ead_xml_records = Ead.parse_source_file(
         "tests/fixtures/ead/ead_record_attribute_and_subfield_variations.xml"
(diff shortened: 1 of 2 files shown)


tests/sources/xml/test_ead.py
```diff
@@ -1,10 +1,61 @@
+# ruff: noqa: E501
```
ehanson8, line 1: Perhaps just a personal opinion but I think it's preferable to not ignore E501 for a module (or even a line really). For the `xml_string` var in `create_ead_source_record_stub`, you can use spaces instead of tabs and you can break up the long citation strings like this:

```
assert Ead.get_citation(source_record) == (
        "Charles J. Connick Stained Glass Foundation Collection, "
        "VC-0002, box X. Massachusetts Institute of Technology, "
        "Department of Distinctive Collections, Cambridge, Massachusetts."
    )
```

tests/sources/xml/test_ead.py
```diff
@@ -501,12 +563,11 @@ def test_ead_record_invalid_date_and_date_range_are_omitted(caplog):
             assert date.range.lte != "1989"
             assert date.range.gte != "2001"
             assert date.range.lte != "1999"
-
     assert ("has a date that couldn't be parsed: 'undated'") in caplog.text
     assert ("has a later start date than end date: '2001', '1999'") in caplog.text
 
 
-def test_ead_record_correct_identifiers_from_multiple_unitid(caplog):
+def test_ead_transform_with_multiple_unitid_gets_valid_ids():
```
ehanson8, line 570: `Ead` has more of these edge case tests than the other sources we've done so far, we might consider redoing these with stub records but not sure if that's worth the effort

tests/sources/xml/test_ead.py
```diff
@@ -515,19 +576,223 @@ def test_ead_record_correct_identifiers_from_multiple_unitid(caplog):
...
+                        <emph>
+                            Foundation
+                            <emph>Collection</emph>
+                        </emph>
+                        <num>VC.0002</num>
+                    </unittitle>
+                    <unittitle></unittitle>
+                    <unittitle></unittitle>
+                </did>
+            </archdesc>
```
ehanson8, line 626: This is duplicated a lot,  you might refactor so top level elements like `<archdesc>` aren't repeated in every test

ghukill (changes requested):
Overall, I think it looks great!   Left a request for docstrings in the helper methods, and a question, but overall looking good to me.

tests/sources/xml/test_ead.py
```diff
@@ -515,19 +576,223 @@ def test_ead_record_correct_identifiers_from_multiple_unitid(caplog):
...
+        metadata_insert=(
+            """
+            <archdesc level="collection"></archdesc>
+            """
+        )
+    )
+    assert Ead.get_citation(source_record) is None
+
+
+def test_get_main_titles_success():
```
ehanson8, line 697: This would be easier to read if the order matched the method order in `ead.py` so I would move the `get_main_titles` tests further down

transmogrifier/sources/xml/ead.py
```diff
@@ -362,19 +327,86 @@ def generate_name_identifier_url(cls, name_element: Tag) -> list | None:
...
+    def _get_collection_description(cls, source_record: Tag) -> Tag:
+        if collection_description := source_record.metadata.find(
+            "archdesc", level="collection"
+        ):
+            return collection_description
+        message = (
+            "Record skipped because key information is missing: "
+            '<archdesc level="collection">.'
+        )
+        raise SkippedRecordEvent(message)
```
ehanson8, line 339: Perfect use of `SkippedRecordEvent`!

transmogrifier/sources/xml/ead.py
```diff
@@ -17,72 +18,36 @@
...
-            pass
-        else:
-            message = (
-                f"Record ID {self.get_source_record_id(xml)} is missing archdesc element"
-            )
-            logger.error(message)
-            return None
+        # <archdesc> and <did> elements are required when deriving optional fields
+        collection_description = self._get_collection_description(source_record)
+        collection_description_did = self._get_collection_description_did(source_record)
```
ehanson8, line 36: Wondering if these calls are still necessary up top, it seems they would raise the `SkippedRecordEvent` whenever they are called by the field methods

transmogrifier/sources/xml/ead.py
```diff
@@ -362,19 +327,86 @@ def generate_name_identifier_url(cls, name_element: Tag) -> list | None:
...
+        ] or None
+
+    @classmethod
+    def get_citation(cls, source_record: Tag) -> str | None:
+        collection_description = cls._get_collection_description(source_record)
+        if (
+            citation_element := collection_description.find("prefercite", recursive=False)
+        ) and (
+            citation := cls.create_string_from_mixed_value(
+                citation_element, " ", ["head"]
```
ehanson8, line 372: We may want to start using named args for this

ehanson8 (commented):
Very good, a few suggestions!

transmogrifier/sources/xml/ead.py
ghukill: I left a comment that might be related: my assumption is that these will be removed once _fully_ refactored to field methods?

ehanson8: Ah, didn't see your comments before my review but that makes sense as a plan!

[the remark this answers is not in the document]
jonavellecuerdo: Ah, yes! I meant to write a note after marking the PR ready for review but forgot. 😅 Thank you for remembering! I added a note in the PR description!

jonavellecuerdo: See [response](https://github.com/MITLibraries/transmogrifier/pull/185#discussion_r1621324261); PR description updated with note. Sorry for the confusion!

transmogrifier/sources/xml/ead.py
jonavellecuerdo: I was considering this but felt conflicted [with @ghukill 's comment on a previous PR](https://github.com/MITLibraries/transmogrifier/pull/177#discussion_r1613894603), but perhaps the guideline @ghukill is only applicable for "methods with single positional arguments" or "when testing"?

In any case, I will update to use named args! I think `create_string_from_mixed_value` is nifty but is difficult to understand just by the name alone.

ghukill: To clarify my comment a bit...

If I'm reading correctly, this is the signature:
```python
def create_string_from_mixed_value(
        cls,
        xml_element: Tag,
        separator: str = "",
        skipped_elements: list[str] | None = None,
    )...
```

So anytime called, I'd expect to see:
  - required, positional argument `xml_element` is not named, but just passed positionally
  - the rest of them, passed as name arguments, because named in the signature
    - _unless_ the defaults are okay, and then okay to omit

Does that mesh with the discussion here?

ghukill: So in this case, I'd expect:

```python
cls.create_string_from_mixed_value(
    citation_element,
    separator = " ",
    skipped_elements = ["head"]
)
```

Open to pushback, but this convention feels like a pretty widely followed convention?

jonavellecuerdo: Hmm, I think my confusion was I took "named arguments" (or keyword arguments) literally. 🤔 Technically, all the arguments are positional since the signature isn't as follows:

```python
def create_string_from_mixed_value(
        cls,
        xml_element: Tag,
        *,
        separator: str = "",
        skipped_elements: list[str] | None = None,
)
```

I'm definitely on board with the proposed convention; just curious if it's more accurate to say: 

> if overriding default positional arguments, use keyword arguments in the method call
?

tests/sources/xml/test_ead.py
jonavellecuerdo: Good idea! On a somewhat related note (and is also why I hesitated to reformat the strings):

When I save on an open XML file, I have a tool in my IDE that reformats the XML. It results in newlines being added throughout the XML file, which then causes the tests to fail. 

@ehanson8 @ghukill Do you think it's worth creating an issue to whitespace/newline-proof the XML transforms? 🤔

transmogrifier/sources/xml/ead.py
ghukill: If I'm reading it right, this is the actual function definition:

```python
@classmethod
def create_string_from_mixed_value(
    cls,
    xml_element: Tag,
    separator: str = "",
    skipped_elements: list[str] | None = None,
) -> str:
   ...
```

Apologies if I'm being imprecise with my words here or earlier, but I would characterize the arguments like:
  * `cls`: required, placeholder, can kind of ignore
  * `xml_element`: required, positional
  * `separator`: because it has equals sign `=`, it's named, optional, with default
  * `skipped_elements`: same as separator

I take your point that _technically_ all the arguments are positional, good point.  But I suppose my position is that _conventionally_ -- and this is open to debate for sure -- if an argument has an equals sign `=` than it's treated as a named argument, and therefore optional with a default value.

This [SO post](https://stackoverflow.com/a/57819001) breaks it down nicely, and exposes that I _have_ been lazy with my terminology.

If we ever do deal with functions that use `*args` or `**kwargs`, then I think it gets more subtle (to your example above), but for most signatures, I feel like it's often positional (no defaults) and then named/keyword (with defaults).  My vote is to
   * a) _not_ use the arg name for positionals
   * b) use the arg name for named/keyword

Thoughts?  or am I just making a mountain out of a mole hill 😅?

jonavellecuerdo: Thank you for clarifying! I accept / understand your definitions and we are on the same page! I do wonder whether we should start using `*` or `/` in our function signatures to more explicitly distinguish positional vs named/keyword arguments but beyond the scope of this PR. 🤓 If I remember correctly, [Ruff throws errors for boolean positional arguments (FBT001)](https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/).

Circling back to original comment: [changes have been made](https://github.com/MITLibraries/transmogrifier/pull/185/commits/de46532bbe2390a48aed2927793e60a130ba0a50).

ehanson8: That feels right to me but I honestly am learning some of these conventions in these PR conversations 🙃

tests/sources/xml/test_ead.py
ghukill: We'd probably want to preserve whitespace inside an element (its text), as this is often intentional by the record creator.  I think it would fundamentally change the nature of the content if we introduce newline characters inside a long string of text for an element.

But reformatting the XML in a way that just moves elements around, but not element text, that seems okay / good.

jonavellecuerdo: I think I'll forego creating an issue for now as the newlines are resulting from me not turning off the XML formatter in my IDE 😅 . 

Circling back to original comment, I removed the syntax to ignore `E501` errors and made updates accordingly!

transmogrifier/sources/xml/ead.py
ghukill: This conversation is the first time I've seen the `/` operator for function/method signatures, and helped clarify for me what the `*` operator does _without_ an immediate following variable name like `*args`.  

This helped me understand it more tacitly:

```python
In [1]: def foo(a, b, *, c="horse"):
   ...:     print(a,b,c)
   ...: 

In [2]: foo(42, 101, 999, 888, c='zebra')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 foo(42, 101, 999, 888, c='zebra')

TypeError: foo() takes 2 positional arguments but 4 positional arguments (and 1 keyword-only argument) were given

In [3]: def bar(a, b, *args, c="horse"):
   ...:     print(a,b,args,c)
   ...: 

In [4]: bar(42, 101, 999, 888, c='zebra')
42 101 (999, 888) zebra
```

I'm definitely not opposed to using `*` and `/` in signatures if we find them useful.  Here is an edge case where I could see it being useful, when you want to force keyword argument or no positional:

```python
def baz(*, a=1):
    pass

In [6]: baz()
1

In [7]: baz(42)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 baz(42)

TypeError: baz() takes 0 positional arguments but 1 was given

In [8]: baz(a=101)
101    
```

Ineresting stuff!  In cases without `*` or `/`, glad we're honing in on some conventions, and curious if we ever run into situations where it would be nice to use them.  I read an article this morning that I found pretty convincing that the main appeal is often APIs, where you want to finely control how functions are called by users potentially not familiar with the library.

tests/sources/xml/test_ead.py
jonavellecuerdo: @ehanson8 For this PR, I'll stick to just renaming the test and will revisit when I work on the `get_identifiers` field method.

tests/sources/xml/test_ead.py
jonavellecuerdo: Updated!

tests/sources/xml/test_ead.py
jonavellecuerdo: @ehanson8 @ghukill Let me know what you think of the updates I made to the [`create_ead_source_record_stub` method](https://github.com/MITLibraries/transmogrifier/pull/185/commits/b8adf19271c53d9f2b0f49d50d8b71ed0c68bd20#diff-4e446208d5381af3faa605c03e15486e64e9af30c4d54e8f40bec821cb87c851R11-R74). It was the only way I could think of simplifying some of the `metadata_insert` arguments throughout the testing module (removes the `<archdesc>` and `<archdesc><did>` elements from the string.

In general, we want to avoid complex functions in the testing module, but maybe this is a simple enough method that it isn't a problem? 🤔

transmogrifier/sources/xml/ead.py
```diff
@@ -362,19 +327,86 @@ def generate_name_identifier_url(cls, name_element: Tag) -> list | None:
         return None
 
     @classmethod
-    def get_main_titles(cls, xml: Tag) -> list[str]:
+    def _get_collection_description(cls, source_record: Tag) -> Tag:
```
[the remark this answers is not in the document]
jonavellecuerdo: Docstrings have been added!

transmogrifier/sources/xml/ead.py
```diff
@@ -362,19 +327,86 @@ def generate_name_identifier_url(cls, name_element: Tag) -> list | None:
...
+        ):
+            return collection_description
+        message = (
+            "Record skipped because key information is missing: "
+            '<archdesc level="collection">.'
+        )
+        raise SkippedRecordEvent(message)
+
+    @classmethod
+    def _get_collection_description_did(cls, source_record: Tag) -> Tag:
```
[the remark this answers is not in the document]
jonavellecuerdo: Docstrings have been added!

tests/sources/xml/test_ead.py
ehanson8: That's totally fine, just food for thought!

tests/sources/xml/test_ead.py
```diff
@@ -9,29 +8,27 @@
 
 
 def create_ead_source_record_stub(
-    header_insert: str = "", metadata_insert: str = ""
+    header_insert: str = "",
+    metadata_insert: str = "",
+    parent_element: Literal["archdesc", "did"] = None,  # noqa: RUF013
```
ehanson8, line 13: This isn't too complex in my mind and saves space in the unit tests but I defer to @ghukill for final approval of this

ehanson8 (approved):
Looks good to me and I'm glad this PR led to some learning for all of us! 🙂

transmogrifier/sources/xml/ead.py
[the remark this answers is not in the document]
ghukill: Thanks!

transmogrifier/sources/xml/ead.py
```diff
@@ -362,38 +329,139 @@ def generate_name_identifier_url(cls, name_element: Tag) -> list | None:
...
     @classmethod
-    def get_main_titles(cls, xml: Tag) -> list[str]:
+    def _get_collection_description(cls, source_record: Tag) -> Tag:
+        """Get element with archival description for a collection.
+
+        For EAD-formatted XML, all of the information about a collection
+        that is relevant to the TIMDEX data model is
+        encapsulated within the <archdesc level="collection"> element.
+        If this element is missing, it suggests that there is a structural
+        error in the record.
```
ghukill, line 339: Perfect, thanks.

ghukill (approved):
Looks good to me!

Thanks again for good discussions.  I'm glad in these early updates, we're hashing out some of these conventions and details.

PR #192, opened 2024-06-06
Field method refactor for Ead transform

jonavellecuerdo:

Field method refactor for transform class Ead (Part 3).

* Add field methods and corresponding unit tests for: `rights`, `related_items`, and `subjects`

2. Run CLI command
   ```
   pipenv run transform -i tests/fixtures/ead/ead_record_all_fields.xml -o output/ead-transformed-records.json -s aspace
   ```

   Output: 

   ```
   2024-06-06 15:43:31,737 INFO transmogrifier.cli.main(): Logger 'root' configured with level=INFO
   2024-06-06 15:43:31,737 INFO transmogrifier.cli.main(): No Sentry DSN found, exceptions will not be sent to Sentry
   2024-06-06 15:43:31,737 INFO transmogrifier.cli.main(): Running transform for source aspace
   2024-06-06 15:43:31,756 WARNING transmogrifier.sources.transformer.get_valid_title(): Record repositories/2/resources/1 has multiple titles. Using the first title from the following titles found: ['Charles J. Connick Stained Glass Foundation Collection', 'Title 2', 'Title 3']
   2024-06-06 15:43:31,779 INFO transmogrifier.cli.main(): Completed transform, total records processed: 1, transformed records: 1, skipped records: 0, deleted records: 0
   2024-06-06 15:43:31,779 INFO transmogrifier.cli.main(): Total time to complete transform: 0:00:00.042270
   ```

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-287

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_ead.py
+++ b/tests/sources/xml/test_ead.py
@@ -191,12 +191,6 @@ def test_ead_transform_with_all_fields_transforms_correctly():
                 relationship="Alternate Format",
             ),
             timdex.RelatedItem(
-                description=(
-                    "Issues of Twilight Zine were separated for library cataloging."
-                ),
-                relationship="Separated Material",
-            ),
-            timdex.RelatedItem(
                 description="MC-0423 James R. Killian Papers",
             ),
             timdex.RelatedItem(
@@ -243,6 +237,12 @@ def test_ead_transform_with_all_fields_transforms_correctly():
                     "Foundation may be found at their website "
                     "(http://www.cjconnick.org/)."
                 ),
+            ),
+            timdex.RelatedItem(
+                description=(
+                    "Issues of Twilight Zine were separated for library cataloging."
+                ),
+                relationship="Separated Material",
             ),
         ],
         rights=[
@@ -514,14 +514,6 @@ def test_ead_transform_with_attribute_and_subfield_variations_transforms_correct
             ),
             timdex.RelatedItem(
                 description="Data with blank head tag",
-                relationship="Separated Material",
-            ),
-            timdex.RelatedItem(
-                description="Data with no head tag",
-                relationship="Separated Material",
-            ),
-            timdex.RelatedItem(
-                description="Data with blank head tag",
             ),
             timdex.RelatedItem(
                 description="Data with no head tag",
@@ -531,6 +523,14 @@ def test_ead_transform_with_attribute_and_subfield_variations_transforms_correct
             ),
             timdex.RelatedItem(
                 description="List data with no head tag",
+            ),
+            timdex.RelatedItem(
+                description="Data with blank head tag",
+                relationship="Separated Material",
+            ),
+            timdex.RelatedItem(
+                description="Data with no head tag",
+                relationship="Separated Material",
             ),
         ],
         subjects=[
@@ -1144,11 +1144,7 @@ def test_get_notes_transforms_correctly_with_multiple_kinds():
             <bioghist>
                 <head>Biographical Note</head>
                 <p>
-            """
-            "Charles J. Connick (1875-1945) was an American "
-            "<emph>stained</emph> glass artist whose work may be found "
-            "in cities all across the United States."
-            """
+                    Charles J. Connick was an American <emph>stained</emph> glass artist.
                 </p>
                 <p>Connick founded his own studio in 1912 in Boston.</p>
             </bioghist>
@@ -1168,11 +1164,7 @@ def test_get_notes_transforms_correctly_with_multiple_kinds():
         ),
         timdex.Note(
             value=[
-                (
-                    "Charles J. Connick (1875-1945) was an American "
-                    "stained glass artist whose work may be found "
-                    "in cities all across the United States."
-                ),
+                "Charles J. Connick was an American stained glass artist.",
                 "Connick founded his own studio in 1912 in Boston.",
             ],
             kind="Biographical Note",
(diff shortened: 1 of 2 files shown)


transmogrifier/sources/xml/ead.py
```diff
@@ -557,6 +493,96 @@ def get_publishers(cls, source_record: Tag) -> list[timdex.Publisher] | None:
             return [timdex.Publisher(name=publication_value)]
         return None
 
+    @classmethod
+    def get_related_items(cls, source_record: Tag) -> list[timdex.RelatedItem] | None:
+        related_items = []
+        collection_description = cls._get_collection_description(source_record)
+
+        for related_item_element in collection_description.find_all(
```
ghukill, line 501: Can't recall what the convention was in this file, but could optionally just iterate over the method call directly.

transmogrifier/sources/xml/ead.py
```diff
@@ -557,6 +493,96 @@ def get_publishers(cls, source_record: Tag) -> list[timdex.Publisher] | None:
...
+            ["altformavail", "relatedmaterial", "separatedmaterial"], recursive=False
+        ):
+            if related_item_element.name == "relatedmaterial":
+                related_items.extend(cls._get_related_material(related_item_element))
+            else:  # noqa: PLR5501
+                if related_item := cls.create_string_from_mixed_value(
```
ghukill, line 507: The lint ignore caught my attention, think it could be simplified to:

```python
if related_item_element.name == "relatedmaterial":
    related_items.extend(cls._get_related_material(related_item_element))
elif related_item := cls.create_string_from_mixed_value(
    related_item_element, separator=" ", skipped_elements=["head"]
):
```

transmogrifier/sources/xml/ead.py
```diff
@@ -557,6 +493,96 @@ def get_publishers(cls, source_record: Tag) -> list[timdex.Publisher] | None:
             return [timdex.Publisher(name=publication_value)]
         return None
 
+    @classmethod
+    def get_related_items(cls, source_record: Tag) -> list[timdex.RelatedItem] | None:
```
ghukill, line 497: Reading top down, you'll likely encounter this comment before the following which came first.  But I must admit, this method broke my brain.

I had thought perhaps some of the if/else logic could be removed by targeting `<relatedmaterial>` in a list comprehension, and then the others the same way, but looking at the XML, I see now that _all_ values in the test `test_get_related_items_transforms_correctly_with_related_materials` are coming from two `<relatedmaterial>` elements.

I have no doubt it's a faithful copying of pre-existing logic, and probably not worth a rework here, but this was hard to parse for me.  That said, maybe that's just a "me" problem, and kudos to the test for thwarting what I thought were some easy tweaks!

transmogrifier/sources/xml/ead.py
```diff
@@ -557,6 +493,96 @@ def get_publishers(cls, source_record: Tag) -> list[timdex.Publisher] | None:
...
+                            description=related_item,
+                            relationship=aspace_type_crosswalk.get(
+                                related_item_element.name, related_item_element.name
+                            ),
+                        )
+                    )
+        return related_items or None
+
+    @classmethod
+    def _get_related_material(
```
ghukill, line 521: Last comment on these related methods: I still don't know how, or if, there is a more straight-forward way to parse these from those wild semantic + display XML elements, but noting these feel like areas that could be cleaned up a bit.  

I think, for me, this refactor work is nice because it exposes some tricky corners like this, but are likely tricky for good reason because the logic is intricate.  I don't see any need to address now, and maybe not ever, but I'm appreciating having some insight into corners of these transformations I was not familiar with previously.

ghukill (approved):
Nothing blocking from me, so setting as approved.

Did leave a couple comments around "related materials", but mostly just observations and discussion.

transmogrifier/sources/xml/ead.py
ehanson8: FWIW, I've generally been iterating over a variable when a `.find_all(...)` or other call is needed for iteration , `for related_item_element in collection_description.find_all()` rather than `for related_item_element in cls._get_collection_description(source_record).find_all()`

transmogrifier/sources/xml/ead.py
ehanson8: Yes ☝️

transmogrifier/sources/xml/ead.py
ehanson8: Agreed this should have an issue for later review, it's rough (coming from the likely author 🙃 who knows better now)

ehanson8: Taking another look though, could it skip the `related_materials = []` and  just return:

```
return [
                timdex.RelatedItem(description=related_item)
                for subelement in subelements
                if (
                    related_item := cls.create_string_from_mixed_value(
                        subelement, separator=" ", skipped_elements=["head"]
                    )
                )
            ]
```

Apologies if I'm going down the rabbit hole you already went down @ghukill !

ehanson8 (commented):
A few comments!

transmogrifier/sources/xml/ead.py
ghukill: Tis a rabbit hole indeed!  When I saw the XML getting parsed, I immediately appreciated some of the logic here.  It's not easy to extract.

Your proposal above makes sense to me!  Does this mean that elements `["altformavail", "separatedmaterial"]` aren't needed?  and if so, is this true across all EAD records?

As an aside, this is precisely where I think many-record analysis could be extremely helpful.  If we could:
  * run against 100 EADs, save output
  * make change, run against same files, save output
  * have a convenient way to see the difference for this particular field

Lots of approaches come to mind, but the many-record comparison pattern itself is what I'm interested in.

ehanson8: Agree on the utility of record analysis in a situation like this! I was just referring to updating the `_get_related_material` method, I believe `"altformavail` and `separatedmaterial` would still be caught in the updated `elif` block you recommended for `get_related_items`. But this is also a sign that this should be flagged in an issue for later consideration since it's complicated!

transmogrifier/sources/xml/ead.py
jonavellecuerdo: Ah, yes! I've been doing the same as @ehanson8. If it is followed by a `.find`/`.findall`, assign to a variable! Feels a little easier to read.

transmogrifier/sources/xml/ead.py
jonavellecuerdo: Ah, I mistakenly thought that the statement for the `elif` condition would also execute if it was also met, so I hesitated to collapse the `if/else` conditions. This has been updated!

transmogrifier/sources/xml/ead.py
jonavellecuerdo: @ehanson8 Made the change to remove `related_materials = []` from `_get_related_material()` (and directly return the result from list comprehension instead. 

@ghukill Many-record comparison sounds ideal! I think the `attributes_and_subfield_variations` fixture tries to address this, but it takes looking at the XML file to see the different conditions it tries to trigger. I definitely think a convenient way to see the outputs across many records and comparing different runs would be helpful for complex transforms!

Also, I tried to provide more context regarding the updates to `get_related_material` here https://github.com/MITLibraries/transmogrifier/pull/192#discussion_r1631457455! A long read but hoped to provide more clarification!

transmogrifier/sources/xml/ead.py
ghukill: Appreciate the detailed breakdown!  All tracks and makes sense.

At the end of the day, this block has appeared to do what it needs to do, so I'm definitely good with it as-is even.  As dicussed here, and elsewhere, it's tricky given the variable nature of EAD.

If you _did_ want to break out into more private, sub-methods, I might propose going pretty hard in that direction, like:
  * `_get_related_from_relatedmaterial()`
  * `_get_related_from_altformavail()`
  * `_get_related_from_separatedmaterial()`

Perhaps this is a throwback to a comment I made on one of @ehanson8's PRs about a variable name that smushed up two words.  I'm appreciating now that it was probably because the element was named as such, so it feels accurate; and I agree.  Apologies @ehanson8 if I misssed that nuance at the time!  I feel as though, now, I would read these methods -- particularly when they were right next to one another when being called -- and understand their naming is reflective of the XML elements, e.g.

```python
related_items = []
related_items.extend(_get_related_from_relatedmaterial())
related_items.extend(_get_related_from_altformavail())
related_items.extend(_get_related_from_separatedmaterial())
```

But... to your points @jonavellecuerdo about other similar methods not breaking it out into sub-methods like this, I think that's valid.  

I've been appreciating during these PR reviews of the thoughtful consideration to the goal of the refactoring to fields methods vs general refactoring of logic that has not been revisited in awhile.  I would vote for continuing to err on the side of _not_ making too many sweeping changes during this first pass.  Which is not to say changes that have been made were a bad idea, quite the contrary they all seemed good, just that if it gets thorny without a clear, swift improvement, maybe better to just leave for now.  Just my two cents.

jonavellecuerdo: Please see updated tests! Also decided to rename private method to `_get_related_from_relatedmaterial`. Although long, I concur that it is a great naming convention that clearly describes the XML element source!

tests/sources/xml/test_ead.py
```diff
@@ -1144,11 +1144,7 @@ def test_get_notes_transforms_correctly_with_multiple_kinds():
             <bioghist>
                 <head>Biographical Note</head>
                 <p>
-            """
-            "Charles J. Connick (1875-1945) was an American "
-            "<emph>stained</emph> glass artist whose work may be found "
-            "in cities all across the United States."
-            """
+                    Charles J. Connick was an American <emph>stained</emph> glass artist.
```
ghukill, line 1147: Good call on the `<emph>`!  As we chatted about in our call, no real hard and fast rules about what to include, but the more we mimick stuff we might see, the better.  Dig it.

tests/sources/xml/test_ead.py
```diff
@@ -1319,32 +1311,48 @@ def test_get_related_items_success():
...
-            "A use copy of photographic plates in box 4 can be found "
-            "in the Institute Archives and Special Collections reading room.</p>"
-            """
+                <p>A use copy of photographic plates in box 4.</p>
             </altformavail>
             <separatedmaterial>
                 <head>Separated Materials</head>
                 <p>Issues of Twilight Zine were separated for library cataloging.</p>
             </separatedmaterial>
-            """
```
ghukill: Again, ➕ to dropping the triple quotes, and sorry for missing on prior reviews.  Glad we're establishing some patterns now as we continue into others (though as @ehanson8 pointed out, EAD is probably the most garish XML wise).

ghukill (approved):
Approved!  Thanks for the flexibility and revisions.

ehanson8 (approved):

PR #193, opened 2024-06-11
Add final set of DspaceMets field methods

ehanson8:
Finish refactoring `DspaceMets` transform to use field methods. Just under our suggested limit 😬 

Run the following command to see that the `DspaceMets` transform still transforms a source file:

```
pipenv run transform -i tests/fixtures/dspace/dspace_mets_records.xml -o output/dspace-mets-transformed-records.json -s dspace
```

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-286

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_dspace_mets.py
+++ b/tests/sources/xml/test_dspace_mets.py
@@ -4,13 +4,15 @@ import transmogrifier.models as timdex
 from transmogrifier.sources.xml.dspace_mets import DspaceMets
 
 
-def create_dspace_mets_source_record_stub(xml_insert: str = "") -> BeautifulSoup:
+def create_dspace_mets_source_record_stub(
+    dmdsec_insert: str = "", filesec_insert: str = ""
+) -> BeautifulSoup:
     xml_string = f"""
         <records>
          <record xmlns="http://www.openarchives.org/OAI/2.0/"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
           <header>
-            <identifier>abc123</identifier>
+            <identifier>oai:dspace:abc123</identifier>
           <header>
           <metadata>
            <mets xmlns="http://www.loc.gov/METS/"
@@ -24,11 +26,14 @@ def create_dspace_mets_source_record_stub(xml_insert: str = "") -> BeautifulSoup
                xsi:schemaLocation="http://www.loc.gov/mods/v3
                http://www.loc.gov/standards/mods/v3/mods-3-1.xsd">
                 <mods:mods>
-                {xml_insert}
+                {dmdsec_insert}
                 </mods:mods>
               </xmlData>
              </mdWrap>
             </dmdSec>
+            <fileSec>
+            {filesec_insert}
+            </fileSec>
            </mets>
           </metadata>
          </record>
@@ -210,7 +215,7 @@ def test_dspace_mets_with_attribute_and_subfield_variations_transforms_correctly
 
 def test_get_alternate_titles_success():
     source_record = create_dspace_mets_source_record_stub(
-        """
+        dmdsec_insert="""
         <mods:titleInfo>
          <mods:title type="alternative">A Slightly Different Title</mods:title>
         </mods:titleInfo>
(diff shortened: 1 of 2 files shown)


transmogrifier/sources/xml/dspace_mets.py
```diff
@@ -212,6 +136,136 @@ def get_content_type(cls, source_record: Tag) -> list[str] | None:
...
+                        kind=kind,
+                        value=str(name.string),
+                    )
+                )
+        return contributors or None
+
+    @classmethod
+    def get_dates(cls, source_record: Tag) -> list[timdex.Date] | None:
+        # Only publication date is mapped from DSpace, other relevant date field (dc.
+        # coverage.temporal) is not mapped to the OAI-PMH METS output.
```
ehanson8, line 159: There are several of these comments that were copied over, not sure how useful they really are and they feel like clutter to me but wanted to preserve them for your review. There's also this [ticket](https://mitlibraries.atlassian.net/browse/TIMX-16) which could render this entire transform unnecessary should we decide to go with `DspaceDim` for the `DSpace@MIT` instance

ghukill: Ignoring for the moment if we keep `DSpaceMETS`, I think they'd make nice docstrings vs standalone comments.  To me, this is a big benefit of the field methods: a place for this kind of documentation.  [Here's an example from GeoHarvester](https://github.com/MITLibraries/geo-harvester/blob/main/harvester/records/formats/fgdc.py#L291-L297).  The rub is that our linting requires a first line title (short, ending in period) and then can have freeform text.  My two cents is that a fairly generic title like "Field method: <field name>" is okay, then can launch into commentary.

I think it's nice to keep them for now, but thanks for sharing the ticket, that it may be moot eventually.

transmogrifier/sources/xml/dspace_mets.py
```diff
@@ -212,6 +136,136 @@ def get_content_type(cls, source_record: Tag) -> list[str] | None:
...
+                    timdex.Date(kind="Publication date", value=publication_date_value)
+                ]
+        return None
+
+    @classmethod
+    def get_file_formats(cls, source_record: Tag) -> list[str] | None:
+        # Only maps formats with attribute use="ORIGINAL" because other formats such as
+        # USE="TEXT" are used internally by DSpace and not made publicly available.
+        file_formats = []
+        for file_group in source_record.find_all("fileGrp", USE="ORIGINAL"):
```
ghukill, line 175: Does `USE` need to be uppercase?  Not a big deal, just surprising to see it that way.

tests/sources/xml/test_dspace_mets.py
```diff
@@ -4,13 +4,15 @@
 from transmogrifier.sources.xml.dspace_mets import DspaceMets
 
 
-def create_dspace_mets_source_record_stub(xml_insert: str = "") -> BeautifulSoup:
+def create_dspace_mets_source_record_stub(
+    dmdsec_insert: str = "", filesec_insert: str = ""
```
ghukill, line 8: I think this is a nice adjustment of the stub to accomodate more sections.  This feels like a perfect way to incrementally adjust complexity of the stub _as needed_, and not before.

transmogrifier/sources/xml/dspace_mets.py
ghukill: Nevermind!  Keeping comment for historical reviewing purposes, but I see now that the attribute `USE` is uppercase in the DSpace METS.

ghukill (approved):
I think it looks great, approved.

Now that we've had a decent amount of these PRs coming through, I feel like I'm having some secondary level thoughts about transforms, possible helper methods that could be applied across _lots_ of sources and transforms, BS4 vs Xpath/lxml, etc., but feeling like those might be discussions for future tickets or PRs.

What I am seeing is logic faithfully transferred from imperative code to field methods, keeping the complexity of change low, which will keeps confidence high we're not fundamentally changing the transform outputs, just the architecture of the code.  Which, I think is exactly the scope of this work at this time.

transmogrifier/sources/xml/dspace_mets.py
ehanson8: Good call, I shifted them over!

jonavellecuerdo (approved):

PR #194, opened 2024-06-17
Add 1st set of DspaceDim field methods

ehanson8:
Refactors the first set of `DspaceDim` fields as separate field methods. 

Run the following command to see that the `DspaceDim` transform still transforms a source file and skips an invalid content type according to the `Whoas` transform:

```
```

NO

NO

* https://mitlibraries.atlassian.net/browse/TIMX-282

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_dspace_dim.py
+++ b/tests/sources/xml/test_dspace_dim.py
@@ -1,5 +1,33 @@
+from bs4 import BeautifulSoup
+
 import transmogrifier.models as timdex
 from transmogrifier.sources.xml.dspace_dim import DspaceDim
+
+
+def create_dspace_dim_source_record_stub(xml_insert: str = "") -> BeautifulSoup:
+    xml_string = f"""
+        <records>
+         <record xmlns="http://www.openarchives.org/OAI/2.0/"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+          <header>
+           <identifier>oai:darchive.mblwhoilibrary.org:1912/2641</identifier>
+           <datestamp>2020-01-28T19:30:01Z</datestamp>
+           <setSpec>com_1912_3</setSpec>
+           <setSpec>col_1912_534</setSpec>
+          </header>
+          <metadata>
+           <dim:dim xmlns:dim="http://www.dspace.org/xmlns/dspace/dim"
+           xmlns:doc="http://www.lyncode.com/xoai"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://www.dspace.org/xmlns/dspace/dim
+           http://www.dspace.org/schema/dim.xsd">
+           {xml_insert}
+           </dim:dim>
+          </metadata>
+         </record>
+        </records>
+        """
+    return BeautifulSoup(xml_string, "xml")
 
 
 def test_dspace_dim_transform_with_all_fields_transforms_correctly():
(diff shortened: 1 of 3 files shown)


transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -260,47 +183,174 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+
+    @classmethod
+    def _get_creators(cls, source_record: Tag) -> Iterator[timdex.Contributor]:
+        for creator in source_record.find_all(
+            "dim:field", element="creator", string=True
+        ):
+            yield timdex.Contributor(
+                value=str(creator.string),
+                kind="Creator",
+            )
```
ghukill, line 254: Just curious (and same for `_get_contributors...()` below), why the `yield`?

On one hand, I kind of like it: no need for list comprehensions, just yield `Contributor` objects as they are found.  On the other hand: feels a bit different than other places?

FWIW, I do kind of like the `yield`?  Wonder -- not pressing by any means -- if opportunity to move in this direction for other methods as we encounter them?

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -260,47 +183,174 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+        ] or None
+
+    @classmethod
+    def get_contributors(cls, source_record: Tag) -> list[timdex.Contributor] | None:
+        contributors = []
+        contributors.extend(list(cls._get_creators(source_record)))
+        contributors.extend(
+            list(cls._get_contributors_by_contributor_element(source_record))
+        )
```
ghukill, line 243: Related to the comment below asking about use of `yield`, I think we could drop the `list(...)` to convert the generator returns from `_get_creators()` and `_get_contributors...()`. 

I had to test to confirm, but it looks like `<list>.extend()` will hydrate the results of the generator if passed (which kind of makes sense if you think about it, as it's modifying the original list with values):

```python
In [3]: def foo():
   ...:     for x in range(0,10):
   ...:         yield x
   ...: 

In [4]: y = []

In [5]: y.extend(foo())

In [6]: y
Out[6]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

Learn something new everyday!  Don't think it hurts to have the `list(...)` here, but just noting.

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -260,47 +183,174 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+        alternate_titles = [
+            timdex.AlternateTitle(
+                value=str(alternate_title.string),
+                kind=alternate_title["qualifier"],
+            )
+            for alternate_title in source_record.find_all(
+                "dim:field", element="title", string=True
+            )
+            if alternate_title.get("qualifier")
+        ]
```
ghukill, line 198: Whether we ultimately keep BS4, or move into Xpath/lxml, I'm finding this pattern increasingly easy to read.

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -260,47 +183,174 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+                ):
+                    yield timdex.Date(
+                        range=timdex.DateRange(
+                            gte=gte_date,
+                            lte=lte_date,
+                        ),
+                        kind="coverage",
+                    )
+            else:
+                yield timdex.Date(note=coverage_value, kind="coverage")
```
ghukill, line 309: Ah, more `yields`!  Maybe I just missed them in other transforms.  Digging this pattern.

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -260,47 +183,174 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+            date_value = str(date.string.strip())
+            if validate_date(date_value, cls.get_source_record_id(source_record)):
+                if date.get("qualifier") == "issued":
+                    date_object = timdex.Date(value=date_value, kind="Publication date")
+                else:
+                    date_object = timdex.Date(
+                        value=date_value, kind=date.get("qualifier") or None
+                    )
+                dates.append(date_object)
+        dates.extend(list(cls._get_coverage_dates(source_record)))
```
ghukill, line 281: See comment above: could probably drop the `list(...)` here for list extension.

tests/sources/xml/test_dspace_dim.py
```diff
@@ -216,3 +244,231 @@ def test_dspace_dim_transform_with_optional_fields_missing_transforms_correctly(
...
+        timdex.Contributor(value="Jamerson, James", kind="Creator"),
+        timdex.Contributor(
+            value="LaFountain, James R.",
+            kind="author",
+        ),
+        timdex.Contributor(
+            value="Oldenbourg, Rudolf",
+            kind="author",
+        ),
+    ]
```
ghukill, line 381: Noting now, but don't think we should make these changes in this refactor, but this feels like a good opportunity at some point to normalized `kind` values.  Seeing `Creator` (title cased) and `author` (lowercase).  I know `author` comes from the XML itself, and it's always risky to programatically convert case.

Maybe normalization is some kind of post-processing / post record transform?  Again, not needed now, just noting for the future.

transmogrifier/sources/xml/whoas.py
```diff
@@ -23,18 +24,17 @@ class Whoas(DspaceDim):
...
+        content_types = [
+            str(content_type.string)
+            for content_type in source_record.find_all(
+                "dim:field", element="type", string=True
+            )
         ] or ["no content type in source record"]
+        if cls.valid_content_types(content_types):
+            return content_types
+        message = f'Record skipped based on content type: "{content_types}"'
+        raise SkippedRecordEvent(message, cls.get_source_record_id(source_record))
```
ghukill, line 37: 💯

ghukill (approved):
All in all, I think it looks great!  Left a couple comments about yielding + list extending, but I'm finding these really easy to parse and reason about.  Nice work on the changes, and the PR chunking.

transmogrifier/sources/xml/dspace_dim.py
ehanson8: Interesting, I never reconsidered this pattern but glad you pointed it out! I'll remove `list(...)`

tests/sources/xml/test_dspace_dim.py
ehanson8: I was almost certain we had a Jira ticket for this but we don't, I'll create an issue

ehanson8:
@ghukill Had to add some type hinting in `get_contributors` after removing `list(...)

tests/sources/xml/test_dspace_dim.py
ghukill: Thanks for creating the ticket.  Maybe when this field method refactor is done, before we launch into orchestration, we could meet and survey the tickets created?  Get a feel for anything we may want to also tackle during the orchestration?  or after?

ehanson8: Before sounds like a good idea!

ehanson8:
I prefer it without `list(...)` so I'm fine with a occasional inter-method type hinting as a trade-off. I'll clean up the whole repo in a separate commit in my next PR

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -260,47 +183,172 @@ def get_optional_fields(self, xml: Tag) -> dict | None:
...
+                "dim:field",
+                element="description",
+                qualifier="tableofcontents",
+                string=True,
+            )
+        ] or None
+
+    @classmethod
+    def get_contributors(cls, source_record: Tag) -> list[timdex.Contributor] | None:
+        contributors: list[timdex.Contributor] = []
```
ghukill, line 239: I like it.  I'm also a little surprised that mypy needed this to understand what what was coming from those methods which are both clearly typed.

transmogrifier/sources/xml/dspace_dim.py
ghukill: Yeah, those might make for a couple of nice PRs when the first pass is over: "Normalize yield..",  "Normalize BS4 string casting...", etc.  Though I think by and large things are pretty good.

ghukill (approved):
LGTM 👍

PR #196, opened 2024-06-26
Timx 282 dspacedim fmr judgment day

ehanson8:
Finish refactoring `DspaceDim` class to use field methods.

Run the following command to see that the `DspaceDim` transform still transforms a source file:

```
```

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-282

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_dspace_dim.py
+++ b/tests/sources/xml/test_dspace_dim.py
@@ -472,3 +472,308 @@ def test_get_file_formats_transforms_correctly_if_fields_missing():
 
 def test_get_format_success():
     assert DspaceDim.get_format() == "electronic resource"
+
+
+def test_get_funding_information_success():
+    source_record = create_dspace_dim_source_record_stub(
+        """
+        <dim:field mdschema="dc" element="description" qualifier="sponsorship"
+        >NSF Grant Numbers: OCE-1029305, OCE-1029411, OCE-1249353</dim:field>
+        """
+    )
+    assert DspaceDim.get_funding_information(source_record) == [
+        timdex.Funder(
+            funder_name="NSF Grant Numbers: OCE-1029305, OCE-1029411, OCE-1249353",
+        )
+    ]
+
+
+def test_get_funding_information_transforms_correctly_if_fields_blank():
+    source_record = create_dspace_dim_source_record_stub(
+        '<dim:field mdschema="dc" element="description" qualifier="sponsorship" />'
+    )
+    assert DspaceDim.get_funding_information(source_record) is None
+
+
+def test_get_funding_information_transforms_correctly_if_fields_missing():
+    source_record = create_dspace_dim_source_record_stub()
+    assert DspaceDim.get_funding_information(source_record) is None
+
+
+def test_get_identifiers_success():
+    source_record = create_dspace_dim_source_record_stub(
+        """
+        <dim:field mdschema="dc" element="identifier" qualifier="uri">https://hdl.handle.net/1912/2641</dim:field>
+        """
+    )
+    assert DspaceDim.get_identifiers(source_record) == [
+        timdex.Identifier(value="https://hdl.handle.net/1912/2641", kind="uri")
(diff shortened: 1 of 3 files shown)


tests/sources/xml/test_dspace_dim.py
```diff
@@ -472,3 +472,308 @@ def test_get_file_formats_transforms_correctly_if_fields_missing():
...
+
+def test_get_related_items_success():
+    source_record = create_dspace_dim_source_record_stub(
+        """
+        <dim:field mdschema="dc" element="relation"
+        >A low resolution version of this movie was published.</dim:field>
+        <dim:field mdschema="dc" element="relation" qualifier="ispartofseries"
+        >International Association of Aquatic and Marine Science</dim:field>
+        <dim:field mdschema="dc" element="relation" qualifier="uri"
+        >https://doi.org/10.1002/2016JB013228</dim:field>
```
ghukill, line 658: This is a very minor nit, not a required or even suggested change, but I think personally I would find these kind of stubs easier to read if the elements were one line?  

e.g.,
```python
source_record = create_dspace_dim_source_record_stub(
    """
    <dim:field mdschema="dc" element="relation">A low resolution version of this movie was published.</dim:field>
    <dim:field mdschema="dc" element="relation" qualifier="ispartofseries">International Association of Aquatic and Marine Science</dim:field>
    <dim:field mdschema="dc" element="relation" qualifier="uri">https://doi.org/10.1002/2016JB013228</dim:field>
    """
)
```

I understand this breaks line length linting... but the readability might be worth it sometimes.

Again, not requesting a change -- particularly as it's probably fairly common across transforms -- but noting to propose that it's an option when we want.

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -320,6 +223,146 @@ def get_file_formats(cls, source_record: Tag) -> list[str] | None:
...
+        return [
+            timdex.Identifier(
+                value=str(identifier.string),
+                kind=identifier.get("qualifier") or "Not specified",
+            )
+            for identifier in source_record.find_all(
+                "dim:field", element="identifier", string=True
+            )
+            if identifier.get("qualifier") != "citation"
+        ] or None
```
ghukill, line 248: Using this as an example, but I'm finding these easy to scan now.  I do wonder if it's because I've seen quite a few now: list comprehension + filtering clause + `or` for some values.  It's complex, but I think once you take the time to understand one, you kind of understand the rest in the transform.

When we get into orchestration, and all FMR is done, it might be worth looking at some of these and see if some helper utilities can be constructed.

ghukill (approved):
Looks great to me, nice work.  I left a couple of conversational comments, but nothing requested.

tests/sources/xml/test_dspace_dim.py
ehanson8: We can talk about that, not a big fan of sidescrolling, but not sure if I'm the only one so a later discussion is warranted!

transmogrifier/sources/xml/dspace_dim.py
ehanson8: Yes! I agree that this pattern is ripe for further abstraction. Created an [issue](https://github.com/MITLibraries/transmogrifier/issues/197) for this!

tests/sources/xml/test_dspace_dim.py
jonavellecuerdo: I think I prefer side-scrolling 🤔. If the XML tags themselves are long, I would also prefer:

```
<tag> 
   [some long string]
</tag>
```

jonavellecuerdo (approved):
Looks good to me!

tests/sources/xml/test_dspace_dim.py
ehanson8: Got it, yes, let's discuss further!

Issue #188, opened 2024-05-31
Revisit logging of date parsing errors


ghukill:
Two immediate things come to mind when it comes to parsing / validating dates with `validate_date`, `validate_date_range`, etc.:
1. we are forced to pull the `source_record_id` to pass it to the function, such that it can _maybe_ be used for logging
2. how helpful are the log messages themselves?

It has been discussed here and there, but probably worth revisiting these helper functions and think about how we can more seamlessly use them in code.

PR #198, opened 2024-07-10
Timx 285 aardvark fmr

ehanson8:
Refactors the first set of `Aardvark` field methods to align with the conventions developed since this class was initially created.

Run the following command to see that the `Aardvark` class transform still transforms a source file:

```
pipenv run transform -o output/gismit.json -i  tests/fixtures/aardvark_record_all_fields.jsonl -s gismit
```

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-285

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/json/test_aardvark.py
+++ b/tests/sources/json/test_aardvark.py
@@ -147,14 +147,54 @@ def test_aardvark_record_is_deleted_returns_false_if_value_is_false(
     assert MITAardvark.record_is_deleted(source_record) is False
 
 
-def test_aardvark_get_alternate_titles_success(aardvark_record_all_fields):
-    assert MITAardvark.get_alternate_titles(next(aardvark_record_all_fields)) == [
+def test_aardvark_get_alternate_titles_success(aardvark_records):
+    source_record = next(aardvark_records)
+    source_record["dct_alternative_sm"] = ["Alternate title"]
+    assert MITAardvark.get_alternate_titles(source_record) == [
         timdex.AlternateTitle(value="Alternate title")
     ]
 
 
-def test_aardvark_get_contributors_success(aardvark_record_all_fields):
-    assert MITAardvark.get_contributors(next(aardvark_record_all_fields)) == [
+def test_aardvark_get_alternate_titles_transforms_correctly_if_fields_blank(
+    aardvark_records,
+):
+    source_record = next(aardvark_records)
+    source_record["dct_alternative_sm"] = []
+    assert MITAardvark.get_alternate_titles(source_record) is None
+
+
+def test_aardvark_get_alternate_titles_transforms_correctly_if_fields_missing(
+    aardvark_records,
+):
+    source_record = next(aardvark_records)
+    assert MITAardvark.get_alternate_titles(source_record) is None
+
+
+def test_aardvark_get_content_type_success(aardvark_records):
+    source_record = next(aardvark_records)
+    source_record["gbl_resourceType_sm"] = ["Vector data"]
+    assert MITAardvark.get_content_type(source_record) == ["Vector data"]
+
+
+def test_aardvark_get_content_type_transforms_correctly_if_fields_blank(
+    aardvark_records,
+):
+    source_record = next(aardvark_records)
+    source_record["gbl_resourceType_sm"] = []
+    assert MITAardvark.get_content_type(source_record) is None
+
+
+def test_aardvark_get_content_type_transforms_correctly_if_fields_missing(
+    aardvark_records,
+):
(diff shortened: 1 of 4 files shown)


tests/sources/json/test_aardvark.py
```diff
@@ -147,14 +147,54 @@ def test_aardvark_record_is_deleted_returns_false_if_value_is_false(
     assert MITAardvark.record_is_deleted(source_record) is False
 
 
-def test_aardvark_get_alternate_titles_success(aardvark_record_all_fields):
-    assert MITAardvark.get_alternate_titles(next(aardvark_record_all_fields)) == [
+def test_aardvark_get_alternate_titles_success(aardvark_records):
+    source_record = next(aardvark_records)
```
ehanson8, line 151: Stub records are a lot different with JSON vs. XML so I don't fully approve of this particular approach but figure we would discuss in this PR before I go further. All suggestions welcome so we can properly set our JSON source testing conventions going forward!

transmogrifier/sources/json/aardvark.py
```diff
@@ -178,58 +179,42 @@ def get_optional_fields(self, source_record: dict) -> dict | None:
...
-        issued_dates = []
-        if "dct_issued_s" in source_record:
-            issued_dates.append(
-                timdex.Date(value=source_record["dct_issued_s"], kind="Issued")
-            )
-        return issued_dates
+    def _issued_dates(cls, source_record: dict) -> Iterator[timdex.Date]:
+        if issued_date := source_record.get("dct_issued_s"):  # noqa: SIM102
+            if validate_date(issued_date, cls.get_source_record_id(source_record)):
+                yield (timdex.Date(value=issued_date, kind="Issued"))
```
ehanson8, line 214: I'm shifting the date validation before  the `Date` instance is created, which seems more inline with how we're doing it in other transforms but again, I want to hear everyone's thoughts on this!

tests/sources/json/test_aardvark.py
ghukill: I'd agree that JSON records are easier to manipulate, given that once was parsed it's just a dictionary we can easily modify.

Instead of using the overhead of an iterator and `next(...)`, what about a similar stub method -- for symmetry -- that just returns a barebones dictionary?  Then the test is free to add/remove keys as needed.

While I think a test could just do:
```python
source_record = {}
```

The benefit of a stub might be that we could set some shared defaults, maybe bits and bobs that we know are commonly accessed.  But at the end of the day, it _is_ just a dictionary.

transmogrifier/sources/json/aardvark.py
ghukill: I'm good with this pattern!  This is very easy to parse and understand.  With a nod to the [oustanding date parsing issue](https://github.com/MITLibraries/transmogrifier/issues/188), I think this looks great.

ghukill (approved):
Other than a couple of conversational comments, overall looks great!  Thinks it's a testament to the quality of the original pass.

tests/sources/json/test_aardvark.py
ehanson8: You are correct about the value of symmetry, I'll create a method!

ehanson8:
@ghukill @jonavellecuerdo Pushed a new commit and re-requesting review

tests/sources/json/test_aardvark.py
```diff
@@ -7,6 +7,10 @@
 from transmogrifier.sources.json.aardvark import MITAardvark
 
 
+def create_aardvark_source_record_stub() -> dict:
+    return {"id": "123", "dct_title_s": "Test title 1"}
```
ghukill, line 11: 👍

transmogrifier/sources/json/aardvark.py
```diff
@@ -102,12 +102,12 @@ def record_is_deleted(cls, source_record: dict) -> bool:
...
             source_record: A JSON object representing a source record.
         """
-        if isinstance(source_record["gbl_suppressed_b"], bool):
+        if isinstance(source_record.get("gbl_suppressed_b"), bool):
             return source_record["gbl_suppressed_b"]
 
         message = (
             f"Record ID '{cls.get_source_record_id(source_record)}': "
-            "'gbl_suppressed_b' value is not a boolean"
+            "'gbl_suppressed_b' value is not a boolean or missing"
```
ghukill, line 110: Nice updates here!  Good catch.

ghukill (approved):

jonavellecuerdo (approved):

PR #200, opened 2024-07-22
Timx 288 marc field method refactor

jonavellecuerdo:

Field method refactor for transform class `MARC` (Part 1).

* Added private methods for key elements in MARC-formatted XML records (`<leader>`, `<controlfield tag="008">`(
* Added field methods and corresponding unit tests for `alternate_titles` and `call_numbers`.

Note(s): 
* For the time-being, I've added calls to the private methods at the beginning of `get_optional_fields()`, which is required by field derivations (that will eventually be moved into their own field methods).
* The name `get_control_field_general_info` was chosen over `get_fixed_length_data` as there are technically _multiple_ fixed length fields in MARC. The fixed length field that MARC depends on is for [code 008, which describes "general information" about the record](https://www.loc.gov/marc/bibliographic/bd008.html).

2. Run CLI command
   ```
   pipenv run transform -i tests/fixtures/marc/marc_record_all_fields.xml -o output/marc-transformed-records.json -s alma
   ```
   Output: 
   ```
   2024-07-22 13:23:04,806 INFO transmogrifier.cli.main(): Logger 'root' configured with level=INFO
   2024-07-22 13:23:04,806 INFO transmogrifier.cli.main(): No Sentry DSN found, exceptions will not be sent to Sentry
   2024-07-22 13:23:04,806 INFO transmogrifier.cli.main(): Running transform for source alma
   2024-07-22 13:23:05,330 INFO transmogrifier.cli.main(): Completed transform, total records processed: 1, transformed records: 1, skipped records: 0, deleted records: 0
   2024-07-22 13:23:05,330 INFO transmogrifier.cli.main(): Total time to complete transform: 0:00:00.524199
   ```

NO

NO

https://mitlibraries.atlassian.net/browse/TIMX-288

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/sources/xml/test_marc.py
+++ b/tests/sources/xml/test_marc.py
@@ -1,9 +1,54 @@
+# ruff: noqa: E501, SLF001
 import logging
 
+import pytest
 from bs4 import BeautifulSoup  # type: ignore[import-untyped]
 
 import transmogrifier.models as timdex
+from transmogrifier.exceptions import SkippedRecordEvent
 from transmogrifier.sources.xml.marc import Marc
+
+
+def create_marc_source_record_stub(
+    leader_field_insert: str = "<leader>03282nam  2200721Ki 4500</leader>",
+    control_field_insert: str = (
+        '<controlfield tag="008">170906s2016    fr mun| o         e zxx d</controlfield>'
+    ),
+    datafield_insert: str = "",
+):
+    """
+    Create source record for unit tests.
+
+    Args:
+        leader_field_insert (str): A string representing a MARC fixed length 'leader'
+            XML element. Defaults to a dummy value.
+        control_field_insert (str): A string representing a MARC fixed length
+            'general info control field' (i.e., code 008) XML element.
+            Defaults to a dummy value.
+        datafield_insert (str): A string representing a MARC 'datafield' XML element.
+
+    Note: A source record for "missing" field method tests can be created by
+        setting datafield_insert = "" (the default).
+    """
+    xml_string = """
+        <collection>
+            <record>
+                {leader_field_insert}
+                {control_field_insert}
+                <controlfield tag="001">990027185640106761</controlfield>
+                {datafield_insert}
+            </record>
+        </collection>
+    """
+
+    return BeautifulSoup(
+        xml_string.format(
+            leader_field_insert=leader_field_insert,
+            control_field_insert=control_field_insert,
+            datafield_insert=datafield_insert,
+        ),
+        "xml",
+    )
 
 
 def test_marc_record_all_fields_transform_correctly():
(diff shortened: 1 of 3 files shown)


ghukill:
One question for both @jonavellecuerdo and @ehanson8: I can't comment directly in the PR, but wondering about moving these static file mappings to methods/properties on the `MarcTransformer` class?

https://github.com/MITLibraries/transmogrifier/pull/200/files#diff-58573f6e57860b560075cbe825c45aace3f2830bb51d47f8d8dbf4b1b0b5d5ebR14-R30

Seems like we'd want to avoid them reading the file each time they are used, but that could be solved with cached instances, or even just read and save on class init.

transmogrifier/sources/xml/marc.py
```diff
@@ -32,109 +33,41 @@
...
-            return None
+        source_record_id = Marc.get_source_record_id(source_record)
 
-        leader = xml.find("leader", string=True)
-        if leader is None:
-            message = f"Record ID {source_record_id} is missing MARC leader"
-            logger.error(message)
-            return None
+        leader_field = Marc._get_leader_field(source_record)
+        control_field_general_info = Marc._get_control_field_general_info(source_record)
```
ghukill, line 50: I see your thinking a bit in having these temporarily here for use in field that don't have their own methods yet... but what about just calling these new and/or renamed methods from those places directly?

e.g. I see `leader_field` used, but could call directly:
```python
# content_type
if content_type := Marc.json_crosswalk_code_to_name(
    self._get_leader_field(source_record)[6:7],
    marc_content_type_crosswalk,
    source_record_id,
    "Leader/06",
):
    fields["content_type"] = [content_type]
```

What I think this might help guide is how often these are used.  If often, it might be worth caching the results on those sub-methods.  If not often, it's probably okay to invoke them directly 1-2 times.  I'm a big fan of caching -- where you call the method, and it reuses data if safe to cache and already called -- but there is a bit of overhead there.

Either way, might avoid needing these temporary values at all, and would give some insight into usage during future parts of this refactor.   Just a thought.

ghukill (commented):
Left a question about the temporary values you noted in the PR, asking if they could just be called directly even if not part of a field method proper.

Also, maybe dovetails with a general comment about the static variables at the top.

These all seem to have one thing in common: parse some data that multiple fields will need.  If so, then feels like we could do that on init for the `MarcTransform` class somehow, and reuse that data throughout via `self`.  As the first part of the MARC refactoring, might be a good time to consider that pattern.

tests/sources/xml/test_marc.py
```diff
@@ -775,6 +793,62 @@ def test_marc_record_with_missing_optional_fields_transforms_correctly():
...
+    with pytest.raises(
+        SkippedRecordEvent,
+        match=(
+            'Record skipped because key information is missing: <controlfield tag="008">.'
+        ),
+    ):
+        Marc._get_control_field_general_info(source_record)
+
+
+def test_get_control_field_general_info_if_field_missing():
```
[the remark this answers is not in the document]
ehanson8: Actually looks like we've used `_raises_skipped_record_event` in other transforms

tests/sources/xml/test_marc.py
```diff
@@ -894,11 +968,7 @@ def test_marc_record_missing_leader_logs_error(caplog):
     output_records = Marc("alma", marc_xml_records)
```
ehanson8, line 968: I'd update this test name since it no longer logs an error

tests/sources/xml/test_marc.py
```diff
@@ -908,11 +978,7 @@ def test_marc_record_missing_008_logs_error(caplog):
...
     assert len(list(output_records)) == 0
     assert output_records.processed_record_count == 1
```
ehanson8, line 980: Same here with the test name

ehanson8 (commented):
Great work, a few minor suggestions!

tests/sources/xml/test_marc.py
```diff
@@ -1,17 +1,30 @@
...
+):
     """
     Create source record for unit tests.
 
     Args:
+        leader_field_insert (str): A string representing a MARC fixed length 'leader'
+            XML element. Defaults to a dummy value.
+        control_field_general_info_insert (str): A string representing a MARC fixed length
+            'general info control field' (i.e., code 008) XML element.
+            Defaults to a dummy value.
```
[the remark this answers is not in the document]
jonavellecuerdo: Thank you! Heads up: Renamed `control_field_general_info_insert` -> `control_field_insert`.

tests/sources/xml/test_marc.py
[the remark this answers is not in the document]
jonavellecuerdo: Good catch! I renamed the test as suggested.

tests/sources/xml/test_marc.py
jonavellecuerdo: I updated the test to say `skips_record` for now, but I think these two tests may be part of the old tests we'll want to deprecate eventually!

tests/sources/xml/test_marc.py
jonavellecuerdo: See previous comment above.

jonavellecuerdo:
> One question for both @jonavellecuerdo and @ehanson8: I can't comment directly in the PR, but wondering about moving these static file mappings to methods/properties on the `MarcTransformer` class?
> 
> https://github.com/MITLibraries/transmogrifier/pull/200/files#diff-58573f6e57860b560075cbe825c45aace3f2830bb51d47f8d8dbf4b1b0b5d5ebR14-R30
> 
> Seems like we'd want to avoid them reading the file each time they are used, but that could be solved with cached instances, or even just read and save on class init.

As @ghukill and I discussed, we opted to only make the following change: **assign 'crosswalk' variables as attributes of `Marc` transform class** for the following reasons: 
* Setting them as class attributes has the same effect as having them as "loose" global variables in the module in that they are instantiated only _once_ when the `Marc` module is imported; whereas calling a class / property method for each one would be called for every transformed record.
* The idea of having these variables set on the instance via a `__post_attrs_init__` was also floated, but `transmogrifier` does not currently use `attrs`, this would be a heavier lift. We also spoke about having a custom `__init__` method but realized that the base `Transformer.__init__` method is "final" and can't be overridden.

transmogrifier/sources/xml/marc.py
jonavellecuerdo: Updated `Marc` to call the private methods as needed but as discussed, caching is not needed at this time!

ehanson8:
> * Setting them as class attributes has the same effect as having them as "loose" global variables in the module in that they are instantiated only _once_ when the `Marc` module is imported; whereas calling a class / property method for each one would be called for every transformed record.

Good summary and I agree that seems like the best path!

transmogrifier/sources/xml/marc.py
```diff
@@ -11,28 +11,24 @@
...
+        "config/holdings_format_crosswalk.json", "json"
+    )
+    holdings_location_crosswalk = load_external_config(
+        "config/holdings_location_crosswalk.json", "json"
+    )
+    language_code_crosswalk = load_external_config("config/loc-languages.xml", "xml")
+    marc_content_type_crosswalk = load_external_config(
+        "config/marc_content_type_crosswalk.json", "json"
+    )
+
```
ehanson8, line 31: Perfection!

ehanson8 (approved):

ghukill:
> > One question for both @jonavellecuerdo and @ehanson8: I can't comment directly in the PR, but wondering about moving these static file mappings to methods/properties on the `MarcTransformer` class?
> > https://github.com/MITLibraries/transmogrifier/pull/200/files#diff-58573f6e57860b560075cbe825c45aace3f2830bb51d47f8d8dbf4b1b0b5d5ebR14-R30
> > Seems like we'd want to avoid them reading the file each time they are used, but that could be solved with cached instances, or even just read and save on class init.
> 
> As @ghukill and I discussed, we opted to only make the following change: **assign 'crosswalk' variables as attributes of `Marc` transform class** for the following reasons:
> 
> * Setting them as class attributes has the same effect as having them as "loose" global variables in the module in that they are instantiated only _once_ when the `Marc` module is imported; whereas calling a class / property method for each one would be called for every transformed record.
> * The idea of having these variables set on the instance via a `__post_attrs_init__` was also floated, but `transmogrifier` does not currently use `attrs`, this would be a heavier lift. We also spoke about having a custom `__init__` method but realized that the base `Transformer.__init__` method is "final" and can't be overridden.

Thanks for making this change.  Kind of minor, but a bit of nice code organization I think.  Also, while I was definitely using "loose" in our conversations, it occurred to me those kind of variables are probably most accurately described as "module level" variables.  Maybe that's a good term we can use going forward to have a collective understanding of what we're talking about.

ghukill (approved):
As a part 1, looks great to me!

ehanson8:
> Thanks for making this change. Kind of minor, but a bit of nice code organization I think. Also, while I was definitely using "loose" in our conversations, it occurred to me those kind of variables are probably most accurately described as "module level" variables. Maybe that's a good term we can use going forward to have a collective understanding of what we're talking about.

I like "module level" and am always in favor of standardizing the vocabulary for our coding!

PR #205, opened 2024-08-01
Timx 291 orchestration

ehanson8:
Update the application's orchestration to streamline the code and improve maintainability

Run a JSON and an XML transform to ensure that records are still transformed successfully after the orchestration refactor:

```
```
```
```

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-291

### Developer

### Code Reviewer(s)


transmogrifier/sources/transformer.py
```diff
@@ -242,30 +238,21 @@ def _transform(
             source_record: A single source record.
         """
         if self.record_is_deleted(source_record):
-            source_record_id = self.get_source_record_id(source_record)
-            timdex_record_id = self.get_timdex_record_id(
-                self.source, source_record_id, source_record
-            )
+            timdex_record_id = self.get_timdex_record_id(source_record)
```
ehanson8, line 241: Since it was brought up in an earlier discussion, do we still want a `_transform` method that is called by the other base transformers with accurate type hints or is it worth consolidating around a single method for both?

transmogrifier/models.py
```diff
@@ -216,3 +216,19 @@ class TimdexRecord:
...
     def asdict(self) -> dict[str, Any]:
         return asdict(self, filter=lambda _, value: value is not None)
+
+    @classmethod
+    def get_required_field_names(cls) -> list[str]:
+        return [
+            field
+            for field, attribute in attrs.fields_dict(cls).items()
+            if attribute.default is attrs.NOTHING
+        ]
```
ehanson8, line 226: As discussed, I currently don't see a use for this but open to keeping it if a different approach to orchestration is identified during the review

transmogrifier/sources/transformer.py
```diff
@@ -242,30 +238,21 @@ def _transform(
...
+        )
+        for field_name, field_method in self.get_optional_field_methods():
+            setattr(timdex_record, field_name, field_method(source_record))
 
-        # If citation field was not present, generate citation from other fields
-        if fields.get("citation") is None:
-            fields["citation"] = generate_citation(fields)
-        if fields.get("content_type") is None:
-            fields["content_type"] = ["Not specified"]
+        self.generate_derived_fields(timdex_record)
```
jonavellecuerdo, line 253: Hmm, I think these still need to be assigned to `timdex_record` variables? 🤔

transmogrifier/sources/transformer.py
```diff
@@ -362,70 +328,98 @@ def record_is_deleted(cls, source_record: dict[str, JSON] | Tag) -> bool:
...
+        self, timdex_record: timdex.TimdexRecord
+    ) -> timdex.TimdexRecord:
         """
-        return {}
+        Generate field values based on existing values in TIMDEX record.
+
+        May not be overridden.
+        """
+        self.create_dates_and_locations_from_publishers(timdex_record)
+        self.create_locations_from_spatial_subjects(timdex_record)
```
jonavellecuerdo, line 355: I think these two also need to be assigned to a `timdex_record` variable.

transmogrifier/sources/transformer.py
```diff
@@ -362,70 +328,98 @@ def record_is_deleted(cls, source_record: dict[str, JSON] | Tag) -> bool:
...
+
+        if timdex_record.citation is None:
+            timdex_record.citation = generate_citation(timdex_record)
+        if timdex_record.content_type is None:
+            timdex_record.content_type = ["Not specified"]
+
+        return timdex_record
```
jonavellecuerdo, line 362: In our discussions, we've referred to these steps are "post transform"/"post processing" and am wondering if the new function can be renamed to include these phrases? 🤔

transmogrifier/sources/xml/dspace_dim.py
```diff
@@ -14,77 +14,6 @@
...
-        # rights
-        fields["rights"] = self.get_rights(source_record)
-
-        # subjects
-        fields["subjects"] = self.get_subjects(source_record)
-
-        # summary
-        fields["summary"] = self.get_summary(source_record)
-
-        return fields
```
jonavellecuerdo, line 86: Huzzah! 🎉

transmogrifier/sources/xml/datacite.py
```diff
@@ -255,7 +187,7 @@ def get_file_formats(cls, source_record: Tag) -> list[str] | None:
         ] or None
 
     @classmethod
-    def get_format(cls) -> str:
+    def get_format(cls, _source_record: Tag) -> str:
```
jonavellecuerdo, line 190: I would've totally missed this being a required change. Good catch!

transmogrifier/sources/transformer.py
ghukill: I'm still torn about this, but I think it'd be nice to collapse to a single method on the base `Transformer` class.  Seems like if we did `source_record: Tag | dict[str, JSON]`, then each time we do `source_record` in the signatures of the field methods for the actual transforms, they could just set the correct type?

transmogrifier/sources/transformer.py
```diff
@@ -242,30 +238,21 @@ def _transform(
...
 
-        fields = self.create_dates_and_locations_from_publishers(fields)
-        fields = self.create_locations_from_spatial_subjects(fields)
+        timdex_record = timdex.TimdexRecord(
+            source=self.source_name,
+            source_link=self.get_source_link(source_record),
+            timdex_record_id=self.get_timdex_record_id(source_record),
+            title=self.get_valid_title(source_record),
+        )
+        for field_name, field_method in self.get_optional_field_methods():
```
ghukill, line 250: Loving this use of `get_optional_field_methods()`.  We've talked about this quite a bit, but this looks and feels really nice.

transmogrifier/sources/transformer.py
```diff
@@ -242,30 +238,21 @@ def _transform(
...
+            setattr(timdex_record, field_name, field_method(source_record))
```
ghukill, line 251: What happens when a field method fails?  Seems pretty clear that we don't want to continue with the record, so I think the question is do we want to catch the exception(s) here, or let them bubble up to the `__next__()` method that called this `transform()` method.

ghukill: Maybe it's a combination of the two?

I see in `__next__()` we have this pattern:
```python
source_record = next(self.source_records)
self.processed_record_count += 1
try:
    record = self.transform(source_record)
except DeletedRecordEvent as error:
    self.deleted_records.append(error.timdex_record_id)
    continue
except SkippedRecordEvent:
    continue
if not record:
    continue
return record
```

It feels correct that we'd let `DeletedRecordEvent` or `SkippedRecordEvent` bubble up from `transform()` to this level, as it's at the "loop" level of transforming.  

So if a field method has an issue, maybe we want to `try/except` it at the `setattr()` level in `transform()`, log the error (and maybe send metrics in the future...) and then manually raise a `SkipRecord...` exception?

transmogrifier/sources/transformer.py
jonavellecuerdo: Hmm, just to clarify-- @ehanson8 , are you asking about the `get_timdex_record_id` field method?

From my understanding, what we currently do is:
* `Transformer.get_timdex_record_id(source_record: Tag | dict[str, JSON])`
* `JSONTransformer.get_timdex_record_id(source_record: dict[str, JSON])`
* `XMLTransformer.get_timdex_record_id(source_record: Tag)`

For me, I think this is understandable!

jonavellecuerdo (commented):
Some initial comments!

tests/sources/xml/test_datacite.py
```diff
@@ -609,7 +609,7 @@ def test_get_file_formats_transforms_correctly_if_fields_missing():
 
 
 def test_get_format_success():
-    assert Datacite.get_format() == "electronic resource"
+    assert Datacite.get_format({}) == "electronic resource"
```
jonavellecuerdo, line 612: Hmm, rather than passing in an empty dictionary, can we set a default of `None` to `_source_record` parameter for the field methods that don't use source records? 🤔 Would change the function signature very slightly:

```
def get_format(cls, _source_record: Tag | None = None) -> str:
        return "electronic resource"
```

```

File: transmogrifier/sources/transformer.py [hunk 11]
```diff
@@ -362,70 +328,98 @@ def record_is_deleted(cls, source_record: dict[str, JSON] | Tag) -> bool:
...
+            if hasattr(self, f"get_{field_name}")
+        ]:
+            yield field_name, getattr(self, f"get_{field_name}")
 
-        Args:
-            source_record: A single source record.
+    @final
+    def generate_derived_fields(
+        self, timdex_record: timdex.TimdexRecord
+    ) -> timdex.TimdexRecord:
```
ghukill, line 348: This method gives me a little pause, though I feel like it's conceptually what we want: some kind of operation that writes field values, shared by all transformers, based on values set by the transformers.

What feels kind of unique, unexpected, is that it **modifies** the `timdex_record` instance directly.  By comparison:
1. for required fields, we see them explicitly set during `timdex_record` init
2. for optional fields, we use `setattr()` to set fields based on the outputs of field methods

What if... this method _yielded_ values that we want to set in the timdex record?  If so, we could set them in a "visible" fashion much like we do for optional fields.  And theoretically, we could also add similar `try/except` logic for each.

For the optional fields loop we know the field name and get a value, so can set it:
```python
for field_name, field_method in self.get_optional_field_methods():
    setattr(timdex_record, field_name, field_method(source_record))
```

What if we did something similar for "dynamic" fields?
```python
for field_name, field_value in self.generate_derived_fields(timdex_record):
    setattr(timdex_record, field_name, field_value)
```

This would mean this method generate_derived_fields() would yield field name + field value each time it determines it should.

Zooming all the way back to `transform()`, I feel like it sets up a record building pattern:
1. init record with required fields
2. build up with optional
3. finalize with dynamic

ghukill: Not sure if this would be easier to test, but it feels like it might?  Any given part of `generate_derived_fields` we could test with a `list(...)` and see the values we _would_ set on the `TimdexRecord`.

transmogrifier/models.py
ghukill: I'd still vote to keep it.  To me the value is symmetry (as you mentioned), and leaning into separation of concerns.  If we imagine for a moment we have no idea how these will be used, they still have values for unambiguously telling us which fields are requierd, and which are optional, for a TIMDEX record.   Having only optional fields odd in that context.

Of course, it turns out maybe we only use that method, but from the lens of the `TimdexRecord` class only, I think it makes good sense to have both.

transmogrifier/sources/transformer.py
ehanson8: > But if a field method has an issue, maybe we want to `try/except` it at the `setattr()` level in `transform()`, log the error (and maybe send metrics in the future...) and then manually raise a `SkipRecord...` exception?

That approach feels good to me and seems to fit with the generalized `SkippedRecordEvent` exception.

transmogrifier/sources/transformer.py
ehanson8: Re: @ghukill's comment below, this should be refactored. Even though it works, it's incongruent with the patterns we've been using so far

transmogrifier/sources/transformer.py
ehanson8: I forgot to add a comment about this: as I thought more about this, I think the derived fields should be considered part of the `transform`. The transformed record isn't really "done" until we add the derived final so that transform/post-transform split doesn't feel as important as when we first discussed it. 

I think it's important to keep it as a separate method from the required/optional fields loop but that method still belongs in `transform` in my mind. Regardless of my views though, I also want to hear both of your thoughts on this!

tests/sources/xml/test_datacite.py
ehanson8: Very good call, much cleaner, thanks!

transmogrifier/sources/transformer.py
ehanson8: Totally correct, this breaks the pattern and the optional fields pattern is very appropriate, the `source_record` would just become the `timdex_record` in these methods. And yes it will be easier to test, the tests I had for this were more involved than necessary

ehanson8: The only complication vs. `optional_fields` is that we need to check if `dates` and `locations` were already set and which would determine whether we `setattr` for new attributes or `extend` the existing attributes. That shouldn't be too difficult though because I was already doing it in the first draft of these methods

ehanson8: Deduping is another issue, this is starting to look more complicated than I initially hoped

transmogrifier/sources/transformer.py
```diff
@@ -351,8 +351,21 @@ def generate_derived_fields(
...
+
+        derived_locations: list[timdex.Location] = []
+        derived_locations.extend(self.create_locations_from_publishers(timdex_record))
+        derived_locations.extend(
+            self.create_locations_from_spatial_subjects(timdex_record)
+        )
+        if not timdex_record.locations:
+            timdex_record.locations = derived_locations if derived_locations else None
+        else:
+            timdex_record.locations.extend(derived_locations)
```
ehanson8, line 372: Very open to suggestions on how to improve this block!

tests/sources/test_transformer.py
ehanson8, line 1: `transformer.py` is getting increasingly difficult to navigate as it gets longer so open to ideas on how to break it up or make it more manageable in some other way

transmogrifier/sources/transformer.py
```diff
@@ -248,7 +248,11 @@ def _transform(
             title=self.get_valid_title(source_record),
         )
         for field_name, field_method in self.get_optional_field_methods():
-            setattr(timdex_record, field_name, field_method(source_record))
+            try:
+                setattr(timdex_record, field_name, field_method(source_record))
+            except:  # noqa: E722
```
ehanson8, line 253: Not sure if we want to narrow this bare exception block and I'm also not sure the best way of testing that an exception triggers this behavior, I did some experiments with the already-imported `Datacite` class but had no luck

ghukill (commented):
Apologies for organizing these comments as a "review".  Submitting a review with comments for discussion here.

transmogrifier/sources/transformer.py
```diff
@@ -362,70 +328,98 @@ def record_is_deleted(cls, source_record: dict[str, JSON] | Tag) -> bool:
...
 
-        May be overridden by source subclasses.
+        May not be overridden.
+        """
+        for field_name in [
+            field_name
+            for field_name in timdex.TimdexRecord.get_optional_field_names()
+            if hasattr(self, f"get_{field_name}")
+        ]:
+            yield field_name, getattr(self, f"get_{field_name}")
```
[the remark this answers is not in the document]
ehanson8: Great suggestion, much more readable!

transmogrifier/sources/transformer.py
ehanson8: Elegant refactor, thanks!

transmogrifier/models.py
jonavellecuerdo: Let's keep it! As @ghukill , I think it's a useful method when considering the `TimdexRecord` class (i.e., at any point, we can call `TimdexRecord.get_required_field_names`).

transmogrifier/sources/transformer.py
jonavellecuerdo: Thank you both for explaining! It helps to view it this way, and I agree, `generate_derived_fields` feels part of the `transform` compared to performing validation or metrics.

transmogrifier/sources/transformer.py
jonavellecuerdo: What is the result of this `try/except` block when the `field_method(source_record)` raises a `SkippedRecordEvent` (e.g., `Datacite.get_content_type`)?

transmogrifier/models.py
ehanson8: Sounds good!

transmogrifier/sources/transformer.py
ehanson8: I believe it would catch `SkippedRecordEvent`, log it and raise it again, does that seem right @ghukill? It's a double dip on the exception in that case but perhaps not a big problem if it is catching all other exceptions and ensuring the record is skipped rather than tanking the whole run

tests/sources/test_transformer.py
ehanson8: Let's talk about this at our planning meeting later this week!

transmogrifier/sources/transformer.py
ehanson8: @ghukill @jonavellecuerdo Surfacing this thread in this increasingly labyrinthian draft PR, do we want to go with a single `Transformer.transform` method or continue with the `Transformer._transform` that is called by the `JSONTransformer` and `XMLTransformer` with more specific type hints.

I think `Transformer.transform` is preferable but defer to the majority

transmogrifier/sources/transformer.py
jonavellecuerdo: Hmm, is there a particular reason why we don't do the following:

```
try: 
   setattr(timdex_record, field_name, field_method(source_record))
except Exception as exc:
   raise SkippedRecordEvent from exc
```

Here is the output I'm seeing:

```
try:
   raise SkippedRecordEvent("Blah")
except Exception as exc:
   raise SkippedRecordEvent from exc

# Output 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
transmogrifier.exceptions.SkippedRecordEvent: Blah

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
transmogrifier.exceptions.SkippedRecordEvent: None
```

jonavellecuerdo: Another option is shown below, but it feels incorrect to say "another" exception occurred. 🤔 

```
try:
    raise SkippedRecordEvent("Blah")
except Exception as exc:
    raise SkippedRecordEvent(exc)

# Output
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
transmogrifier.exceptions.SkippedRecordEvent: Blah

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
transmogrifier.exceptions.SkippedRecordEvent: Blah
```

jonavellecuerdo: Maybe this is the solution (pending `ruff` saying "nope!" 😅 )

```
try:
   raise SkippedRecordEvent("useful error message")
except Exception as exc:
   raise SkippedRecordEvent(exc) from exc
```

Output: 

```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
transmogrifier.exceptions.SkippedRecordEvent: useful error message

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
transmogrifier.exceptions.SkippedRecordEvent: useful error message
```

ehanson8:
@ghukill @jonavellecuerdo I'm making this PR official, thank you both so very much for all of the discussion!

transmogrifier/sources/transformer.py
ghukill, line 1: Can we remove this comment from the top of the file?  Doesn't seem as relevant anymore...

```
# Note: the lxml module in defusedxml is deprecated, so we have to use the
# regular lxml library. Transmogrifier only parses data from known sources so this
```

transmogrifier/sources/xmltransformer.py
ghukill, line 1: Maybe we can remove this message from this file as well?  And the associated `# nosec B410` on `lxml` import?

```
# Note: the lxml module in defusedxml is deprecated, so we have to use the
# regular lxml library. Transmogrifier only parses data from known sources so this
```

transmogrifier/sources/transformer.py
ghukill, line 1: Can't comment on the line itself, as it hasn't changed, but curious people's thoughts on a re-ordering of methods in the `Transformer` class.  This would reorder after a few months of little changes to it, that I think have slowly made the order of methods a bit confusing.

What feels kind of natural to me:

```
class Transformer:

  def get_transformer()...
  def load()...
  
  def transform()...
  def transform_and_write_output_files()...
  def _write_timdex_records_to_json_file()...
  def _write_deleted_records_to_txt_file()...

  # rest of methods, as-is, in order
```

Don't intend to be nit-picky, but if we're anticipating future changes to this class for Parquet writing, thought it might help to organize the methods roughly in conceptual and called order.

Non-block, non-required, just a suggestion.

transmogrifier/sources/transformer.py
```diff
@@ -229,67 +222,36 @@ def parse_source_file(cls, source_file: str) -> Iterator[dict[str, JSON] | Tag]:
...
-    ) -> timdex.TimdexRecord | None:
+    def transform(self, source_record: dict[str, JSON] | Tag) -> timdex.TimdexRecord:
         """
-        Private method called for both XML and JSON transformations, where
-        all logic is shared except source_record type.
+        Transform source record into TimdexRecord instance.
+
+        Instantiates a TimdexRecord instance with required fields and runs fields methods
+        for optional fields. The optional field methods return values or exceptions that
+        prompt the __next__ method to skip the entire record.
```
ghukill, line 231: Very nice, thanks for adding!

One request: maybe a final sentence explaining that after optional fields are set, some shared-across-sources derived fields are also attempted from required + optional field values set by the source transformer?

transmogrifier/sources/transformer.py
```diff
@@ -229,67 +222,36 @@ def parse_source_file(cls, source_file: str) -> Iterator[dict[str, JSON] | Tag]:
...
-        fields = {
-            **self.get_required_fields(source_record),
-            **optional_fields,
-        }
+        timdex_record = timdex.TimdexRecord(
+            source=self.source_name,
+            source_link=self.get_source_link(source_record),
+            timdex_record_id=self.get_timdex_record_id(source_record),
+            title=self.get_valid_title(source_record),
+        )
```
ghukill, line 247: Continue to like this init.  I find it comforting that if this block of explicit attempts for those required fields fail, this record will fail itself.

transmogrifier/sources/transformer.py
```diff
@@ -362,70 +314,107 @@ def record_is_deleted(cls, source_record: dict[str, JSON] | Tag) -> bool:
...
+    def get_optional_field_methods(self) -> Iterator[tuple[str, Callable]]:
+        """
+        Return optional TIMDEX field names and corresponding methods.
+
+        May not be overridden.
+        """
+        for field_name in timdex.TimdexRecord.get_optional_field_names():
+            field_method = getattr(self, f"get_{field_name}", None)
+            if field_method:
+                yield field_name, getattr(self, f"get_{field_name}")
```
ghukill, line 327: What about:

```python
for field_name in timdex.TimdexRecord.get_optional_field_names():
    if field_method := getattr(self, f"get_{field_name}", None):
        yield field_name, field_method
```

transmogrifier/sources/xmltransformer.py
```diff
@@ -44,46 +42,6 @@ def parse_source_file(cls, source_file: str) -> Iterator[Tag]:
...
-    def transform(self, source_record: Tag) -> timdex.TimdexRecord | None:
-        """
-        Call Transformer._transform method to transform XML record to TIMDEX record.
-
-        May not be overridden.
-
-        Args:
-            source_record: A BeautifulSoup Tag representing a single XML record.
-        """
-        return self._transform(source_record)
```
ghukill, line 57: Nice work on this removal, and subsuming this into `Transformer.transform()` method!  Same for `JSONTransformer`.

transmogrifier/helpers.py
```diff
@@ -1,34 +1,33 @@
...
 from datetime import UTC, datetime
 
+import transmogrifier.models as timdex
 from transmogrifier.config import DATE_FORMATS
 
 logger = logging.getLogger(__name__)
 
 
-def generate_citation(extracted_data: dict) -> str:
+def generate_citation(timdex_record: timdex.TimdexRecord) -> str:
```
ghukill, line 10: Nice changes related to this passing of instantiated `timdex_record` vs `extracted_data` dict. 😎

transmogrifier/sources/xmltransformer.py
ghukill, line 1: File level comment re: `record_is_deleted()` method.

While looking at this block in `Transformer.transform()`, got me thinking about the method `record_is_deleted()` that transformers should have:

```python
if self.record_is_deleted(source_record):
    timdex_record_id = self.get_timdex_record_id(source_record)
    raise DeletedRecordEvent(timdex_record_id)
```

I noticed that `JSONTransformer` has an abstract method that must be defined, while `XMLTransformer` actually has an implementation that is used if not overridden.  The default for `XMLTransformer` appears to be an OAI-PMH opinionation:

```python
@classmethod
    def record_is_deleted(cls, source_record: Tag) -> bool:
        """
        Determine whether record has a status of deleted.

        May be overridden by source subclasses if needed.

        Args:
            source_record: A BeautifulSoup Tag representing a single XML record
        """
        return source_record.find("header", status="deleted") is not None
```

I think this is okay.... but not ideal that:
1. `XMLTransformer`  has a default while `JSONTransformer` has an abstract
2. `XMLTrasnformer.record_is_deleted` is opinionated for OAI-PMH

The edge case I see: I'm new to TIMDEX, I'm creating a new source transform for a non OAI-PMH XML source.  Because I'm not _required_ to define a `record_is_deleted()` method, I don't think about it much, and never solve how to identify deleted records for my source.  I may assume somehow that it's just magically handled, when in fact that's only for OAI-PMH sources.

One possible approach: we create a new `Oai` transformer.  Example transformer class lineage for something like `aspace`:
```
Transformer --> XMLTransformer --> Oai --> Ead
```

In this way, the class `Oai` transformer could define any generic OAI-PMH behavior like identifying deleted records, pulling identifier out of header, get source link, etc., a lot of which are defined for `XMLTransformer`.  

Said another way: the `XMLTransformer` class is heavily OAI-PMH opinionated, breaking some symmetry with JSONTransformer.  If we add an intermediate `Oai` class, we could shuffle those opinionations to that class, and keep `XMLTransformer` on par with `JSONTransformer`.

FWIW, this doesn't seem urgent, and potentially slightly out of scope for changes here.  Perhaps better as a standalone ticket?

ghukill (commented):
Overall, I think this looks superb!  Nice work @ehanson8.  Thanks for taking a complex topic, faciliating discussions around changes, and materializing all that in this PR.  While large, it's easy to follow, and the changes feel in-step with things we've been discussing.

As promised, saved some granular and nit picky questions / comments for this "real" PR.  Functionality-wise, really no concerns: I see this operating as intended, and tests and linting is looking good.  Comments are mostly stylistic, method organization, and some questions that may be better spun off as separate issues.

Curious thoughts on a couple of the questions, but close to approval with or without any changes.

transmogrifier/sources/transformer.py
ehanson8: Good call!

ehanson8: That reorder makes sense to me and would also address some of the issues I was having navigating the class, @jonavellecuerdo your thoughts on this?

transmogrifier/sources/transformer.py
ehanson8: Good call!

transmogrifier/sources/transformer.py
ehanson8: Great suggestion!

transmogrifier/sources/xmltransformer.py
ehanson8: We have discussed this before and I agree that the opinionated `XMLTransformer` is worthy of further examination. I feel this comment can be the ticket or would it be better as a GH issue?

ghukill: @ehanson8 - unsure why, but feels like a Jira ticket?  I must admit, I don't have a firm framework in my head for one vs the other.

ehanson8: Created a ticket and to that point, I'd like to talk about that at a future DataEng meeting because it feels like our practices are still ambiguous on what goes where and when. Adding an agenda item to our meeting topic doc

ghukill: Thanks @ehanson8 !

transmogrifier/sources/transformer.py
jonavellecuerdo: @ghukill I like it! For clarification, do the magic methods `__iter__` and `__next__` fall under the category of "rest of methods, as-is, in order"? 🤔

transmogrifier/sources/transformer.py
jonavellecuerdo: @ehanson8 Has this change been implemented? 🤔

transmogrifier/sources/xmltransformer.py
jonavellecuerdo: 👀 Interesting. Thanks for pointing this out @ghukill . Ticket sounds good and looking forward to discussing!

jonavellecuerdo:
@ehanson8 The additional updates are looking good to me! Waiting for response re: ordering (had a follow-up question for @ghukill ) and think [one change request still needs to be implemented](https://github.com/MITLibraries/transmogrifier/pull/205#discussion_r1705541309).

transmogrifier/sources/transformer.py
ehanson8: Not yet, I was saving the commit to include your suggestions as well!

transmogrifier/sources/transformer.py
ghukill: @jonavellecuerdo - I would propose to keep those at the top, where they currently are.  If for no other reason than it's a little unusual / advanced to override those, so should be one of the first things a reader sees when looking at this class.

ehanson8: That makes sense to me!

ehanson8:
@ghukill @jonavellecuerdo Changes pushed!

jonavellecuerdo (approved):

ghukill (approved):
Make it so!

PR #206, opened 2024-08-09
Timx 332 dedupe function

jonavellecuerdo:

Add a global dedupe method using `attrs` converters to remove duplicated data in `TimdexRecord` fields that comprise of item lists. The following approach was taken:

* Define an `attrs` converter method to dedupe list of items
   * [Converter](https://www.attrs.org/en/stable/init.html#converters) methods can be used to normalize values as they are assigned to an `attrs` class attribute.
   * A [`dedupe` method](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR38-R41) is defined. The method accepts an `item_list` and uses `list(dict.fromkeys(item_list))` to remove duplicates **while retaining the order of items** as they were passed into the list. 
* [Create a `ListField` abstract class with a `__hash__` method](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR52-R62)
   * While the `ListField` was implemented [as we discussed](https://docs.google.com/document/d/1JuKJRmmNK1tdrHYDzwEZf5ZTaChXnLwvJYi9G2nOfmo/edit?usp=sharing), the syntax to have the custom classes inherit the hash method was different than expected. As the [Python 3 docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__) explains:
      > If a class that overrides [__eq__()](https://docs.python.org/3/reference/datamodel.html#object.__eq__) needs to retain the implementation of [__hash__()](https://docs.python.org/3/reference/datamodel.html#object.__hash__) from a parent class, the interpreter must be told this explicitly by setting __hash__ = <ParentClass>.__hash__
      This condition applies as `attrs` overrides / implements `__eq__` methods by default, hence the [hash methods are set for the TIMDEX classes as instructed](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR70).
   * [Why is there a list comprehension in `ListField.__hash__`?](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR56-R61) This is required by the custom classes whose attributes _also_ comprise of list fields (i.e., [Contributor](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR74-R81), [Note](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR161-R163), and [Subject](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587cR197-R199)).

1. Review new tests added in commit: [add dedupe tests for dates and locations](https://github.com/MITLibraries/transmogrifier/pull/206/commits/672dc05982ba7d95cfd32541e5cc00b6195a4d39)

2. Review screenshots showing Docker runs using pulled Docker image from **Dev** (representing latest code in this branch) and Docker image from **Prod** (representing current state). Instructions are provided in the attached Markdown file: [simple_ab_testing.md](https://github.com/user-attachments/files/16625990/simple_ab_testing.md).

image

**Note:** All three of us (@ghukill , @ehanson8 , myself) tried this out on our local machines and got similar results! More formal A/B testing will be explored as a separate effort, but all three of us agreed this method is sufficient for purposes of this PR. 

NO

NO

https://mitlibraries.atlassian.net/browse/TIMX-332

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,15 +1,6 @@
 import pytest
 
-from transmogrifier.models import (
-    AlternateTitle,
-    Contributor,
-    Date,
-    DateRange,
-    Identifier,
-    Link,
-    Note,
-    Subject,
-)
+import transmogrifier.models as timdex
 
 
 def test_timdex_record_required_fields_only(timdex_record_required_fields):
@@ -45,14 +36,16 @@ def test_timdex_record_required_fields_only(timdex_record_required_fields):
 
 
 def test_timdex_record_required_subfields_only(timdex_record_required_fields):
-    timdex_record_required_fields.contributors = [Contributor(value="Smith, Jane")]
-    timdex_record_required_fields.identifiers = [Identifier(value="123")]
+    timdex_record_required_fields.contributors = [timdex.Contributor(value="Smith, Jane")]
+    timdex_record_required_fields.identifiers = [timdex.Identifier(value="123")]
     timdex_record_required_fields.links = [
-        Link(url="http://dx.doi.org/10.1007/978-94-017-0726-8")
+        timdex.Link(url="http://dx.doi.org/10.1007/978-94-017-0726-8")
     ]
-    timdex_record_required_fields.notes = [Note(value=["This book is awesome"])]
-    timdex_record_required_fields.alternate_titles = [AlternateTitle(value="Alt Title")]
-    timdex_record_required_fields.subjects = [Subject(value=["Stuff"])]
+    timdex_record_required_fields.notes = [timdex.Note(value=["This book is awesome"])]
+    timdex_record_required_fields.alternate_titles = [
+        timdex.AlternateTitle(value="Alt Title")
+    ]
+    timdex_record_required_fields.subjects = [timdex.Subject(value=["Stuff"])]
     assert timdex_record_required_fields.source == "A Cool Repository"
     assert timdex_record_required_fields.source_link == "https://example.com/123"
     assert timdex_record_required_fields.timdex_record_id == "cool-repo:123"
@@ -337,8 +330,10 @@ def test_timdex_record_date_range_both_gt_and_gte_raises_error(
         match="range may have a 'gt' or 'gte' value, but not both;",
     ):
         timdex_record_required_fields.dates = [
-            Date(
-                range=DateRange(gt="2019-01-01", gte="2019-01-01", lt="2019-06-30"),
+            timdex.Date(
+                range=timdex.DateRange(
+                    gt="2019-01-01", gte="2019-01-01", lt="2019-06-30"
+                ),
             )
         ]
 
@@ -350,8 +345,10 @@ def test_timdex_record_date_range_both_lt_and_lte_raises_error(
         ValueError, match="range may have a 'lt' or 'lte' value, but not both;"
     ):
         timdex_record_required_fields.dates = [
-            Date(
-                range=DateRange(gt="2019-01-01", lt="2019-06-30", lte="2019-06-30"),
+            timdex.Date(
+                range=timdex.DateRange(
+                    gt="2019-01-01", lt="2019-06-30", lte="2019-06-30"
+                ),
             )
         ]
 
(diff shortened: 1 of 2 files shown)


tests/test_models.py
```diff
@@ -1,15 +1,6 @@
...
-from transmogrifier.models import (
-    AlternateTitle,
-    Contributor,
-    Date,
-    DateRange,
-    Identifier,
-    Link,
-    Note,
-    Subject,
-)
```
jonavellecuerdo, line 12: Allows us to easily use `timdex.<field>` without having to explicitly import each field class, similar to what we do for other test modules.

tests/test_models.py
```diff
@@ -374,3 +371,239 @@ def test_timdex_record_not_a_list_raises_error(timdex_record_required_fields):
...
+
+
+def test_timdex_record_dedupe_alternate_titles(timdex_record_required_fields):
+    timdex_record_required_fields.alternate_titles = [
+        timdex.AlternateTitle(value="My Octopus Teacher"),
+        timdex.AlternateTitle(value="My Octopus Teacher"),
+    ]
+    assert timdex_record_required_fields.alternate_titles == [
+        timdex.AlternateTitle(value="My Octopus Teacher")
+    ]
```
jonavellecuerdo, line 383: In the assertions, lists are used because list _allow_ duplicates; with these assertions, the test confirm that a _single instance_ of an object remains in the list **after** the converter function is applied.

transmogrifier/models.py
```diff
@@ -35,6 +35,12 @@ def list_of(item_type: Any) -> Callable:  # noqa: ANN401
     )
 
 
+def dedupe(item_list: list | Any) -> list | None:  # noqa: ANN401
```
jonavellecuerdo, line 38: The `item_list: list | Any` type hint is to support scenarios where [a non-list type is set for a list field](https://github.com/MITLibraries/transmogrifier/pull/206/files#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968R368-R373). Since converters are applied _before_ validators, the `dedupe` converter is set up to "pass" these values so that the **validator** can effectively handle the values.

transmogrifier/models.py
```diff
@@ -43,11 +49,26 @@ def not_empty(
         raise ValueError(message)
 
 
+@define
+class ListField:
+    def __hash__(self) -> int:
+        """Hash method to create unique identifier for Location objects."""
```
ghukill, line 55: Minor: think it's more generic than just `Location objects`?

transmogrifier/models.py
ehanson8: Smart!

tests/test_models.py
```diff
@@ -374,3 +371,239 @@ def test_timdex_record_not_a_list_raises_error(timdex_record_required_fields):
...
+
+
+def test_timdex_record_dedupe_summary(timdex_record_required_fields):
+    timdex_record_required_fields.summary = [
+        "Mitochondria is the powerhouse of the cell.",
+        "Mitochondria is the powerhouse of the cell.",
+    ]
+    assert timdex_record_required_fields.summary == [
+        "Mitochondria is the powerhouse of the cell."
+    ]
```
ehanson8, line 609: Simple and clear tests!

transmogrifier/models.py
```diff
@@ -43,11 +49,26 @@ def not_empty(
...
+        )
+        return hash(values)
+
+
 @define
 class AlternateTitle:
     value: str = field(validator=instance_of(str))  # Required subfield
     kind: str | None = field(default=None, validator=optional(instance_of(str)))
 
+    __hash__ = ListField.__hash__
```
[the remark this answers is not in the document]
ehanson8: It does feel off to define a class and not use it for inheritance, I think it could be worth a huddle to discuss this further

ehanson8 (commented):
While this is pleasantly concise, I share @ghukill 's concerns about the `ListField.__hash__` pattern so I'd appreciate a chance to discuss further to see if there's a better option!

transmogrifier/models.py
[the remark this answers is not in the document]
ehanson8: And if the class isn't used for inheritance, is it better to just defining a `timdex_object_hash()` function and setting `__hash__ = timdex_object_hash()`? Not fully sure of the implications of that though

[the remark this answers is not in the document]
ghukill: Per [this comment](https://github.com/MITLibraries/transmogrifier/pull/206#discussion_r1714182188) above, I wonder if it's passing because the hash created is _always_ unique, so while no runtime exceptions, it's not actually deduping anything?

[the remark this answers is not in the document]
ehanson8: Circled back to this after doing some more reading (this [thread](https://stackoverflow.com/questions/52390576/how-can-i-make-a-python-dataclass-hashable) was particularly helpful), I think either `unsafe_hash=True` or `timdex_object_hash()` works here, the `ListField` class feels like too much overhead if it can't properly be an inherited class due to the hash issues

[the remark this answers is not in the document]
jonavellecuerdo: Thank you both for looking into this more closely! I agree about the `ListField` class feeling redundant as it's not accomplishing inheritance in the way we expected it to. 🤔 

Just wanted to note a couple items:

Regarding @ghukill 's initial comment: 
* The child classes do not inherit the `__hash__` method from the parent class `ListField` because `attrs` default behavior will write an `__eq__` method for the class. the Python 3 docs, it reads:
   > A class that overrides [__eq__()](https://docs.python.org/3/reference/datamodel.html#object.__eq__) and does not define [__hash__()](https://docs.python.org/3/reference/datamodel.html#object.__hash__) will have its [__hash__()](https://docs.python.org/3/reference/datamodel.html#object.__hash__) implicitly set to None. When the [__hash__()](https://docs.python.org/3/reference/datamodel.html#object.__hash__) method of a class is None, instances of the class will raise an appropriate [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) when a program attempts to retrieve their hash value. . .
   I believe this is why we can't simply inherit the `ListField` class. 🤔 
* `@define(frozen=True)` will make the list field (it decorates) **immutable**. The cause of all those errors cropping up when this setting is configured are field methods that involve creating the instance of the object and then proceeding to update attributes (e.g. [dates field method in Ead transform](https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/sources/xml/ead.py#L257-L263)). Given the current state of our transforms, I agree that this approach is not feasible.

To be continued ...

transmogrifier/models.py
jonavellecuerdo: Ah, typo! 😅 Will update!

transmogrifier/models.py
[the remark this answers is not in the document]
ghukill: I'd like to clarify my comment a bit here after a discussion yesterday with @jonavellecuerdo.

I had not considered in this toy example that different classes (e.g. `DogC` vs `DogD`) _should_ likely not be equal, given they are different classes.  

While I was focused on instance equality vs hash method equality... @jonavellecuerdo observed that the class type should be a consideration.

Example, the following should _not_ be equal:
```python
AlternateTitle(value="x", kind="y") == Location(value="x", kind="y")
```

If we are thinking only of field names + values, then they _would_ be considered equal.  But it's clear that to dedupe, the class type is an important variable.  I got the impression that @jonavellecuerdo was exploring changes to reflect this, so won't dig in more here, but did want to include these noets that stemmed from that conversation yesterday.

transmogrifier/models.py
jonavellecuerdo: Added a standalone `timdex_object_hash` function with docstring!

transmogrifier/models.py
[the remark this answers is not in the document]
jonavellecuerdo: @ghukill Thank you for providing this summary! TLDR: From https://github.com/MITLibraries/transmogrifier/pull/206#discussion_r1716965905, I noticed that what `unsafe_hash=True` did _differently_ (at least prior to the updates introduced https://github.com/MITLibraries/transmogrifier/pull/206/commits/f81475b90ba01921858358d1f3f58a8f556110ef) is that it took _class types_ into account when generating the hash method. 

Prior to the updates in https://github.com/MITLibraries/transmogrifier/pull/206/commits/f81475b90ba01921858358d1f3f58a8f556110ef, this would've occurred:
```python
AlternateTitle(value="x", kind="y") == Location(value="x", kind="y") # returns True!
```
To fix this, the [class type is now taken into account my `timdex_object_hash`](https://github.com/MITLibraries/transmogrifier/pull/206/commits/f81475b90ba01921858358d1f3f58a8f556110ef#diff-62380015f300f79c3d62063c1dffd2c085e0ea5a1cfbf24946e5bb31c36d587c). [Unit tests](https://github.com/MITLibraries/transmogrifier/pull/206/commits/f81475b90ba01921858358d1f3f58a8f556110ef#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968R376-R397) to assert that: [if `x == y`, it must follow that `hash(x) == y` (and vice versa)](https://www.attrs.org/en/stable/hashing.html) have also been added.

jonavellecuerdo:
@ghukill @ehanson8 Please review the updates in the latest commit https://github.com/MITLibraries/transmogrifier/pull/206/commits/f81475b90ba01921858358d1f3f58a8f556110ef and the updated description, detailing the simple A/B test we used for reviewing this PR. Thank you for your reviews!

tests/test_models.py
```diff
@@ -373,6 +373,30 @@ def test_timdex_record_not_a_list_raises_error(timdex_record_required_fields):
...
+def test_timdex_object_hash_diff_if_diff_class():
+    """
+    Asserts that TIMDEX objects of different class types
+    with similar attributes and attribute values will
+    be assigned different hashes and declared not equal.
+    """
+    identifier = timdex.Identifier(value="x", kind="y")
+    alternate_title = timdex.AlternateTitle(value="x", kind="y")
+    assert identifier != alternate_title
+    assert identifier.__hash__() != alternate_title.__hash__()
```
ghukill, line 385: Nice test!  I like how explicit these are to test this pretty important little kernel of code and logic.  

Yes, perhaps we _implicitly_ test this in other ways, but it's nice to have this.

transmogrifier/models.py
```diff
@@ -49,25 +43,42 @@ def not_empty(
...
+def timdex_object_hash(timdex_object: Any) -> int:  # noqa: ANN401
+    """Hash method for TIMDEX objects.
+
+    This method is set as the hash method for TIMDEX objects.
+    The method generates a unique hash using a tuple
+    comprised of the class name and attribute values.
+    By making TIMDEX objects hashable, dedupe methods
+    can be applied to a list of TIMDEX objects.
+    """
+    values = tuple(type(timdex_object).__name__)
```
ghukill, line 55: A couple requests here.  

**First**, I think this approach might actually creating a tuple with the individual characters of the class name, e.g.

```python
from transmogrifier.models import *

timdex_object = AlternateTitle(value='horse',kind='animal')

tuple(type(timdex_object).__name__)
# Out[13]: ('A', 'l', 't', 'e', 'r', 'n', 'a', 't', 'e', 'T', 'i', 't', 'l', 'e')
```

Think this could be resolved via an explicit list, and then converting to tuple:
```python
tuple([type(timdex_object).__name__])
# Out[2]: ('AlternateTitle',)
```

**Second**, any objection with using `X.__class__` vs `type(X)`?

Example:
```python
values = tuple([timdex_object.__class__.__name__])
```

Personally, when we _know_ it's some kind of well-known class instance (like classes we define), I think using `.__class__` is more readable.  I'm having trouble finding supporting documentation or reasoning, where it sounds like `type(X)` and `X.__class__` are mostly interchangeable; [this SO post](https://stackoverflow.com/questions/9610993/python-type-or-class-or-is) being a good example where it's kind of fluid, matter of taste, and some pretty edge-y gotchas.

I tend to think of `type(X)` as when I really don't know anything about an object. But in this case, where we _know_ it's a class instance of some kind, and _we_ wrote the classes, it feels like accessing it directly is kind of nice.  

**TL/DR**: I'd propose we make the first change to avoid a tuple of each character, but the second `type(X)` vs `X.__class__` happy either way.

ghukill: ALL that said -- and should have led with this -- this is a nice and elegant solution to this equality checking!  Nice work.

ghukill (changes requested):
Overall, looks great!  

Requesting a change re: the tuple created from the class name.  Also made another stylistic comment, but happy either way on that one.

tests/test_models.py
ehanson8: Agree these are very good tests, great work!

transmogrifier/models.py
ehanson8: I consider `values = tuple([timdex_object.__class__.__name__])` more readable as well, good suggestion!

ehanson8 (commented):
Excellent work, fully approved after that change!

transmogrifier/models.py
jonavellecuerdo: EGAD 😱

jonavellecuerdo: @ghukill Thank you for catching this erroneous error 🙈 . This has been fixed by https://github.com/MITLibraries/transmogrifier/pull/206/commits/426b185b94ab7194345c1c2a23dbb57b2fa15eb6. I also prefer using `__class__` for the same reason you described: 
> we _know_ it's a class instance of some kind, and _we_ wrote the classes

(Also remembering you suggested this during our check-in yesterday so apologies for forgetting!)

ehanson8 (approved):

ghukill (approved):
Approved!! Nice work on this.  Seems like a great addition.

PR #210, opened 2024-11-19
TIMX 403 - Add "run_id" CLI argument and Transformer attribute

ghukill:

As we move into Transmogrifier writing to a parquet dataset, one important bit of information it will need is the concept of a "run id".  This correlates directly to an "Execution UUID" that every StepFunction invocation produces.  This identifier is then used when writing the records to the parquet dataset, allowing for quick and easy access to records associated with that identifier.

There is a small many-to-one relationship that makes naming a bit awkward: each StepFunction invocation may run Transmogrifier multiple times (e.g. multiple input files).  Each time it invokes Transmogrifier, the same `run_id` would be passed.  This effectively groups the outputs of all Transmogrifier invocations under the same `run_id` partition in the parquet dataset.  The language of this new `run_id` in Transmogrifier is intentionally somewhat high level, indicating it's just an identifier to associate with that invocation of the application.

How this addresses that need:
* Adds new CLI argument `-r / --run-id`
* Transformer gets new attribute `run_id`
* Transformer mints a UUID of a run id is not passed, making the change backwards compatible and inconsequential if a run id is not passed

Side effects of this change:
* Going forward, invocations of Transmogrifier can use the `run_id` as part of the parquet record writing.  Until then, it has no effect.

Until the `run_id` is utilized during writing, it has no effect on the output of records at this time.  See the new unit tests for some examples of when and how it's passed.

NO

NO (not yet)

- https://mitlibraries.atlassian.net/browse/TIMX-403

### Developer

### Code Reviewer(s)

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,3 +1,7 @@
+# ruff: noqa: S108
+
+from unittest import mock
+
 from transmogrifier.cli import main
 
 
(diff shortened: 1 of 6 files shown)


transmogrifier/sources/transformer.py
```diff
@@ -72,6 +78,15 @@ def __next__(self) -> timdex.TimdexRecord:
             self.transformed_record_count += 1
             return record
 
+    def set_run_id(self, run_id: str | None) -> str:
+        """Method to set run_id for Transmogrifier run."""
+        if not run_id:
+            logger.info("explicit run_id not passed, minting new UUID")
+            run_id = str(uuid.uuid4())
```
ghukill, line 85: This, specifically, it what keeps this backwards compatible.  If a `run_id` is not passed via the CLI, which our current StepFunction will not do, a UUID is minted... and then just never used in any meaningful way.

ehanson8 (approved):
Looks good to me!

tests/test_cli.py
```diff
@@ -90,3 +94,49 @@ def test_transform_deleted_records(caplog, runner, tmp_path):
...
+        )
+    assert f"run_id set: '{run_id}'" in caplog.text
+
+
+def test_transform_run_id_argument_not_passed_and_uuid_minted(caplog, runner, tmp_path):
+    caplog.set_level("INFO")
+    with mock.patch(
+        "transmogrifier.sources.transformer.Transformer.transform_and_write_output_files"
+    ) as mocked_transform_and_write:
+        mocked_transform_and_write.side_effect = Exception("stopping transformation")
```
jonavellecuerdo, line 128: Is the purpose of this `side_effect` to force stop the `transform` CLI command after loading the transformer (i.e., running the command until we can confirm that the `run_id` attrib is set? 🤔

jonavellecuerdo (approved):
Looks good to me! Just one question. 🤓

tests/test_cli.py
ghukill: That's correct!  I'm sure there are other ways to do it... but this felt kind of convenient.  

And, I suspect that method `transform_and_write_output_files()` will be removed entirely by the parquet work, so I'm not feeling too precious about tests that interact with it.

PR #233, opened 2025-01-16
TIMX 406 - add provenance data to transformed records

ghukill:

This PR adds the field `timdex_provenance` to the TIMDEX data model.  As outlined in the [original engineering plan](https://mitlibraries.atlassian.net/wiki/spaces/IN/pages/4094296066/Engineering+Plan+Parquet+Datasets+for+TIMDEX+ETL#Record-Provenance), this addition to TIMDEX records has multiple uses:

1. providing quick, human readable information about the record's origins (e.g. `run_date`)
2. embed enough information in the indexed Opensearch record to efficiently find the record in the parquet dataset

As outlined in a [recent `timdex-dataset-api` PR](https://github.com/MITLibraries/timdex-dataset-api/pull/73), we need to know a few things to quickly retrieve a record from the dataset: `[run_date, run_type, timdex_record_id, run_record_offset]` where `timdex_record_id` is actually kind of optional, but helpful for confirmation.  This new field contains this information, which is known briefly during write to the dataset in Transmogrifier.

Run a couple of transformations to establish a toy dataset at `/tmp/dataset-provenance`:
```shell
pipenv run transform \
 -s libguides \
 -i tests/fixtures/dataset/libguides-2024-06-03-full-extracted-records-to-index.xml \
 -o /tmp/dataset-provenance \
 --run-id run-abc-123

 pipenv run transform \
 -s libguides \
 -i tests/fixtures/dataset/libguides-2025-01-09-full-extracted-records-to-index.xml \
 -o /tmp/dataset-provenance \
 --run-id run-def-4456
```

Load the dataset:
```python
import json
from timdex_dataset_api import  TIMDEXDataset

td = TIMDEXDataset('/tmp/dataset-provenance')
td.load()
```

Look at a couple of `timdex_provenance` sections from run `run-abc-123` (3 total records):
```python
for record in td.read_transformed_records_iter(run_id='run-abc-123'):
    print(record['timdex_provenance'])
"""
{'source': 'libguides', 'run_date': '2024-06-03', 'run_id': 'run-abc-123', 'run_record_offset': 0}
{'source': 'libguides', 'run_date': '2024-06-03', 'run_id': 'run-abc-123', 'run_record_offset': 2}
{'source': 'libguides', 'run_date': '2024-06-03', 'run_id': 'run-abc-123', 'run_record_offset': 3}
"""
```
Note the absence of `run_record_offset=1`.  The method `read_transformed_records_iter()` only returns records with a transformed record, where we're expecting to see this provenance object, so it didn't show up.

But `run_record_offset=1` actually does exist in the dataset, just for a record that was `action="delete"`:
```python
td.read_dataframe(columns=['timdex_record_id','action','run_record_offset'], run_id='run-abc-123')
"""
Out[14]: 
          timdex_record_id  action  run_record_offset
0  libguides:guides-175846   index                  0
1  libguides:guides-175849  delete                  1              <-------------
2  libguides:guides-175853   index                  2
3  libguides:guides-175855   index                  3
"""
```

This kind of a nicely demonstrates how the offset is calculated relative to the source record encountered during iteration in Transmogrifier.  If records are skipped, they "consume" an offset number, and it keeps incrementing.  In this way, the offset number is wholly artibrary, but by being present in the dataset _and_ the provenance object, it provides a link.

Observe the way `run_record_offset` is written and incremented per run:
```python
# get dataframe of all records in dataset
df = td.read_dataframe()

# get counts of column run_record_offset
# note that offsets 0-3 are repeated twice, this is because the first run only had
# 4 records written, and so offsets 0-3 were repeated for BOTH runs
df.run_record_offset.value_counts()
"""
Out[11]: 
run_record_offset
0      2   <-----
2      2   <-----
3      2   <-----
1      2   <-----
665    1
      ..
338    1
339    1
340    1
341    1
999    1
Name: count, Length: 1001, dtype: int64
"""
```

YES

NO

- https://mitlibraries.atlassian.net/browse/TIMX-406

### Developer

### Code Reviewer(s)

### DIFF
--- a/transmogrifier/models.py
+++ b/transmogrifier/models.py
@@ -213,6 +213,16 @@ class Subject:
 
 
 @define
+class TimdexProvenance:
+    source: str = field(default=None, validator=instance_of(str))
+    run_date: str = field(default=None, validator=instance_of(str))
+    run_id: str = field(default=None, validator=instance_of(str))
+    run_record_offset: int = field(default=None, validator=instance_of(int))
+
+    __hash__ = timdex_object_hash
+
+
+@define
 class TimdexRecord:
     # Required fields
     source: str = field(validator=instance_of(str))
@@ -289,6 +299,9 @@ class TimdexRecord:
     )
     summary: list[str] | None = field(
         default=None, converter=dedupe, validator=optional(list_of(str))
+    )
+    timdex_provenance: TimdexProvenance | None = field(
+        default=None, validator=optional(instance_of(TimdexProvenance))
     )
 
     def asdict(self) -> dict[str, Any]:
--- a/transmogrifier/sources/transformer.py
+++ b/transmogrifier/sources/transformer.py
@@ -72,6 +72,10 @@ class Transformer(ABC):
         if etl_version == 2:  # noqa: PLR2004
             self.run_data = self.get_run_data(source_file, run_id)
 
+    @property
+    def run_record_offset(self) -> int:
+        return self.processed_record_count - 1
+
     @final
     def __iter__(self) -> Iterator[timdex.TimdexRecord | DatasetRecord]:
         """Iterate over transformed records."""
@@ -118,6 +122,12 @@ class Transformer(ABC):
             try:
                 transformed_record = self.transform(source_record)
                 timdex_record_id = transformed_record.timdex_record_id
+                transformed_record.timdex_provenance = timdex.TimdexProvenance(
+                    source=self.run_data["source"],
+                    run_date=self.run_data["run_date"],
+                    run_id=self.run_data["run_id"],
+                    run_record_offset=self.run_record_offset,
+                )
                 self.transformed_record_count += 1
                 action = "index"
 
@@ -145,6 +155,7 @@ class Transformer(ABC):
                     else None
                 ),
                 action=action,
+                run_record_offset=self.run_record_offset,
                 **self.run_data,
             )
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -36,19 +36,19 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:c81223488607457dacb7829ee0c99803c664593b34a2b0f86c71d421e7c3469a",
-                "sha256:d5671226819f6a78e31b1f37bd60f194afb8203254a543d06bdfb76de7d79236"
+                "sha256:159898f51c2997a12541c0e02d6e5a8fe2993ddb307b9478fd9a339f98b57e00",
+                "sha256:d0ca7a58ce25701a52232cc8df9d87854824f1f2964b929305722ebc7959d5a9"
             ],
             "markers": "python_version >= '3.8'",
-            "version": "==1.35.95"
+            "version": "==1.36.0"
         },
         "botocore": {
             "hashes": [
-                "sha256:a672406f748ad6a5b2fb7ea0d8394539eb4fda5332fc5373467d232c4bb27b12",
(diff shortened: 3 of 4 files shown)


transmogrifier/sources/transformer.py
```diff
@@ -118,6 +122,12 @@ def _etl_v2_next_iter_method(self) -> DatasetRecord:
             try:
                 transformed_record = self.transform(source_record)
                 timdex_record_id = transformed_record.timdex_record_id
+                transformed_record.timdex_provenance = timdex.TimdexProvenance(
```
ghukill, line 125: I had considered locating this in the `Transformer.transform()` method, where the `TimdexRecord` instance is built, but opted not to for two reasons:

1. the `transform` method is still v1 and v2 compatible, whereas this would introduce feature flagging
2. the better of the two reasons, I think it's more associated with the orchestration of the run, the `__next__` iterator, etc., and makes sense in this part of the code vs the `transform` method which feels primarily focused on the metadata fields

transmogrifier/sources/transformer.py
```diff
@@ -118,6 +122,12 @@ def _etl_v2_next_iter_method(self) -> DatasetRecord:
...
+                    source=self.run_data["source"],
+                    run_date=self.run_data["run_date"],
+                    run_id=self.run_data["run_id"],
+                    run_record_offset=self.run_record_offset,
+                )
```
ghukill, line 130: I had considered locating this in the `Transformer.transform()` method, where the `TimdexRecord` instance is built, but opted not to for two reasons:

1. the `transform` method is still v1 and v2 compatible, whereas this would introduce feature flagging
2. the better of the two reasons, I think it's more associated with the orchestration of the run, the `__next__` iterator, etc., and makes sense in this part of the code vs the `transform` method which feels primarily focused on the metadata fields

ehanson8 (approved):
Looks good!

transmogrifier/sources/transformer.py
jonavellecuerdo: @ghukill To clarify, the provenance data will be added to TIMDEX records regardless of `ETL_VERSION`, right? It will be added to records generated by both v1 and v2?

transmogrifier/models.py
```diff
@@ -290,6 +300,9 @@ class TimdexRecord:
     summary: list[str] | None = field(
         default=None, converter=dedupe, validator=optional(list_of(str))
     )
+    timdex_provenance: TimdexProvenance | None = field(
+        default=None, validator=optional(instance_of(TimdexProvenance))
+    )
```
jonavellecuerdo, line 305: Why is this field optional? 🤔 In what cases is this field set to `None`?

jonavellecuerdo (approved):
@ghukill Looks good to me! Just one clarification question.

transmogrifier/models.py
ghukill: Bad answer: if it were a "v1" record, it wouldn't have it.  So this simplifies backwards cmpatibility.

Better answer: it allows for creating the `TimdexRecord` incrementally.  You don't have to _initialize_ a record with it set, which allows for setting that property/field more dynamically as the object moves through it's init and handling.  Even better, we somehow validate required fields during serialization.  But doesn't feel that critical.

transmogrifier/sources/transformer.py
ghukill: Not quite - the _model_ is updated to allow it ([see this related response](https://github.com/MITLibraries/transmogrifier/pull/233#discussion_r1929058408)), but only the v2 feature flagged method `_etl_v2_next_iter_method()` will actually add it to the record.

PR #275, opened 2026-08-05
Use 572 digital collections class

ehanson8:
Adding `DigitalCollections` transformer class to extend `DspaceDim` class based on this [metadata mapping](https://mitlibraries.atlassian.net/wiki/x/BQA6QwE). Feedback from stakeholders has been requested so I may delay merging until they have time to weigh-in.

Review Digital Collections records in our `Dev1` UI. **Note:** Reload the page if you get a 500 error since the `Dev1` goes to sleep if unused.

[Sample search for "medicine" in the Digital Collections tab](https://timdex-ui-pipeline-dev.herokuapp.com/results?q=medicine&tab=digital_collections)

YES

YES - `timdex-pipeline-lambdas` and `timdex-index-manager` will require small updates, PRs coming after these changes are merged

- https://mitlibraries.atlassian.net/browse/USE-572

### Code review
* Code review best practices are documented [here](https://mitlibraries.github.io/guides/collaboration/code_review.html) and you are encouraged to have a constructive dialogue with your reviewers about their preferences and expectations.


transmogrifier/sources/xml/digital_collections.py
```diff
@@ -0,0 +1,82 @@
...
+
+import transmogrifier.models as timdex
+from transmogrifier.sources.xml.dspace_dim import DspaceDim
+
+
+class DigitalCollections(DspaceDim):
+    """Digital Collections transformer class."""
+
+    @classmethod
+    def get_funding_information(cls, _source_record: Tag) -> None:
```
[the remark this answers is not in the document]
ehanson8: Noted but ince the type checker didn't flag this, I will leave as-is

ghukill: I see the resolved comment from Copilot; I stumbled on this myself! 

Not required, but wonder if just updating the return signature could do the trick?  

```python
def get_funding_information(cls, _source_record: Tag) -> list[timdex.Funder] | None:
   return None
```

Passes linting for me this way.  This communicates to me, "intentionally overriding this field method to intentionally return no data."  Which, could also be a nice docstring...

transmogrifier/sources/xml/digital_collections.py
```diff
@@ -0,0 +1,82 @@
...
+            timdex.Note(value=[str(technique.string)], kind="technique")
+            for technique in source_record.find_all(
+                "dim:field", mdschema="vra", element="technique", string=True
+            )
+        ] + [
+            timdex.Note(value=[str(medium.string)], kind="medium")
+            for medium in source_record.find_all(
+                "dim:field", element="format", qualifier="medium", string=True
+            )
+        ] or None
```
ghukill: I'll admit, I find this concatenation of list comprehensions kind of dense.  

Alternatively, could save to variables, then concatenate.  Totally optional.

transmogrifier/sources/xml/digital_collections.py
```diff
@@ -0,0 +1,82 @@
...
+                subject.get("qualifier") or "Subject scheme not provided"
+            ].append(str(subject.string))
+        for worktype in source_record.find_all(
+            "dim:field", mdschema="vra", element="worktype", string=True
+        ):
+            subjects_dict["worktype"].append(str(worktype.string))
+        return [
+            timdex.Subject(value=subject_value, kind=subject_scheme)
+            for subject_scheme, subject_value in subjects_dict.items()
+        ] or None
```
ghukill, line 81: Again, kind of stylistic, but I think some whitespace would be helpful.  e.g.

```python
subjects_dict: defaultdict[str, list[str]] = defaultdict(list)

# generic subjects
for subject in source_record.find_all(
    "dim:field", element="subject", string=True
):
    subjects_dict[
        subject.get("qualifier") or "Subject scheme not provided"
    ].append(str(subject.string))

# VRA worktype subjects
for worktype in source_record.find_all(
    "dim:field", mdschema="vra", element="worktype", string=True
):
    subjects_dict["worktype"].append(str(worktype.string))

return [
    timdex.Subject(value=subject_value, kind=subject_scheme)
    for subject_scheme, subject_value in subjects_dict.items()
] or None
```

transmogrifier/sources/xml/digital_collections.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @classmethod
+    def get_summary(cls, source_record: Tag) -> list[str] | None:
```
ghukill, line 84: Given the importance of `summary`, what do you think about a docstring?

I found it surprising that buried down in the list comprehension with conditionals, we include `description` fields that have no qualifier attribute.

Code is terse and I believe correct, but I could imagine a universe where we wonder how/why summaries are getting set, and we could overlook that this captures `<description qualifier="abstract">` or just `<description>`

ghukill (approved):
Left a couple of comments, but overall looks good to me!   

I'd be remiss if I didn't ask: we're pretty sure we want to go with DIM?  I can appreciate the metadata side of things, but I'm a little bummed that we'll have the more minimal DIM files vs METS which do seem to carry more information about the object (surprisingly).  Either way, that's likely a question/discussion out of scope for this PR.  I can live with DIM for sure.  

While I did leave some comments, I'd like to go ahead with an approval review seeing as I'm out Thursday/Friday of this week and don't want to hold things up.  Can always take more passes or revisit things, but everything here looks good _in situ_.

transmogrifier/sources/xml/digital_collections.py
ehanson8: That is what Copilot suggested, I'll make the change to be explicit!

transmogrifier/sources/xml/digital_collections.py
ehanson8: That would be more readable, great suggestion!

transmogrifier/sources/xml/digital_collections.py
ehanson8: Good call, more whitespace let's the code breathe!

transmogrifier/sources/xml/digital_collections.py
ehanson8: Another great suggestion, thanks!

jonavellecuerdo (approved):
Looks good to me! :)

PR #277, opened 2026-08-12
TIMX 655 - handle deletes for researchdatabases (AZ)

ghukill:

**Why these changes are being introduced:**

It turns out that the Springshare OAI endpoint that we harvest AZ items (research databases) has *never* produced deletes.  Without those deleted records / tombstones in OAI, we were not successfully removing records from TIMDEX.

Transmogrifier is currently the place in the TIMDEX ETL ecosystem where records are first written to the dataset, and most commonly, if they are records to index or delete.  There may come a time when explicit pre-transform work is performed to establish source records in the TIMDEX dataset, then we transform them, but Transmogrifier has historically been responsible for that double duty and continues to be at this time.

Ultimately, we need to identify AZ items that are no longer publicly available and establish `action=delete` records in the TIMDEX dataset to have them removed.  The OAI harvester is less per-source opinionated than Transmogrifier, making it a poor choice for this.

**How this addresses that need:**

A new ResearchDatabases transformer class has been created.

Formerly, the TIMDEX source `researchdatabases` used the Springshare transformer class as a naive OAIDC XML transformation.  This new class changes nothing about the metadata transformation, but allows for a place to identify records that were formerly indexed in TIMDEX but are no longer publicly accessible.

There is precedence here in the `libguides` source, which also performs some additional work via the Springshare API.  With that scaffolding already present, it was a relatively simple addition to have `researchdatabases` do a bit of extra work beyond the OAI XML records provided.

The `ResearchDatabases` transformer does something unique: while yielding the OAI XML records provided by the harvester like normal, it also queries the Springshare API to get the current list of public AZ item identifiers and queries the TIMDEX dataset for current `researchdatabases` records.  Any identifiers present in the TIMDEX dataset but absent from the current Springshare API results are no longer public, and for these a *synthetic* OAI XML record is injected into the records yielded by this class for transformation.  Those are handled by pre-existing logic and ultimately get written to the TIMDEX dataset with `action=delete`, prompting a removal from Opensearch downstream.

This approach — querying the TIMDEX dataset for "previous" known identifiers — replaces an earlier design that relied on a managed text file to track identifiers between runs.  Querying the dataset directly avoids the need for an additional file to manage and keeps the source of truth in one place.

Lastly, delete detection is optional.  If any of the required env vars [`LIBGUIDES_API_TOKEN`, `LIBGUIDES_CLIENT_ID`, `TIMDEX_DATASET_LOCATION`] are not set, delete detection is skipped and the transformer behaves as a plain OAI XML transformation, keeping the source fully backwards compatible.

I created two records in the dev TIMDEX dataset for `researchdatabases` that were not present in the Springshare API response (nonsense identifiers).  As such, the following [full StepFunction ETL run in Dev](https://222053980223-2dze743y.us-east-1.console.aws.amazon.com/states/home?region=us-east-1#/v2/executions/details/arn:aws:states:us-east-1:222053980223:execution:timdex-ingest-dev:b4b5734a-f623-4e54-a839-a95b3134c5b3), when running Transmogrifier, [picked up these records should be deleted](https://222053980223-2dze743y.us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/timdex-transmogrifier-dev/log-events/transmogrifier%2Ftimdex-transmogrifier%2Fff24fd53c42a42f2ad81ab4eacf56421):

```
2026-08-12 18:25:30,915 INFO transmogrifier.sources.xml.researchdatabases._yield_api_records_for_deleting(): 2 identifiers identified for deletion: [None, 'i-am-old-record']
...
...
2026-08-12 18:25:35,206 INFO transmogrifier.cli.main(): Completed transform, total records processed: 938, transformed records: 936, skipped records: 0, deleted records: 2
```

_NOTE: you can ignore the `None` in `[None, 'i-am-old-record']`, that was just as mistake mocking some records._

The actual synthetic source records injected into the TIMDEX dataset, via Transmogrifier, looked like this:

```xml
<?xml version="1.0" ?>
<record>
  <header status="deleted">
    <identifier>oai:libguides.com:az/i-am-old-record</identifier>
    <datestamp>2026-08-12T18:25:30Z</datestamp>
    <setSpec>az</setSpec>
  </header>
  <note>This is a synthetic delete record created by Transmogrifier.</note>
  <metadata/>
</record>
```

This synthetic source record is helpful for understanding why/what/how the record is deleted, but it's important to focus that just the presence of a row in the dataset with `action=delete` was sufficient for the rest of the ETL process to remove it from Opensearch.  

NO

YES
* The `researchdatabases` source now requires the Springshare API
credentials and `TIMDEX_DATASET_LOCATION` env var to enable delete
detection, though these are optional for backwards compatibility
* If AZ items are deleted or hidden, they should get removed from
TIMDEX now.

- https://mitlibraries.atlassian.net/browse/TIMX-655

### Code review
* Code review best practices are documented [here](https://mitlibraries.github.io/guides/collaboration/code_review.html) and you are encouraged to have a constructive dialogue with your reviewers about their preferences and expectations.


ehanson8 (approved):
While this lack of deleted records is unfortunate, I like your solution for it, great work! One test naming suggestion

